PR #11, opened 2022-11-29
multiple sources at one bus and no other devices

m-bossart:
This PR addresses another uncommon use case for re-distributing P/Q after running the power flow when there are multiple sources at a bus and no other devices. In this case, there is a check that the source set points match the power flow solution.

The behavior for all other scenarios should remain unchanged.


jd-lara:
We need a test for this, since it is uncommon we need to make sure that this isn't broken later.

jd-lara (approved):

PR #15, opened 2023-02-23
update to PSY2

jd-lara:
This is a preliminary PR to update to PSY 2.0. It will need some additional work to incorporate StandardLoads correctly

### DIFF
--- a/src/post_processing.jl
+++ b/src/post_processing.jl
@@ -130,7 +130,6 @@ function _get_total_p(l::PSY.StandardLoad)
     return PSY.get_constant_active_power(l)
 end
 
-
 function _get_total_q(l::PSY.StandardLoad)
     return PSY.get_constant_reactive_power(l)
 end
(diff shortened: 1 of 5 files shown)


src/post_processing.jl
```diff
@@ -118,17 +118,33 @@ function _update_branch_flow!(sys::PSY.System)
     end
 end
 
+function _get_total_p(l::PSY.PowerLoad)
```
rodrigomha, line 121: These are temporary methods to handle both Standard and Power Loads?

test/test_nlsolve_powerflow.jl
```diff
@@ -86,10 +86,11 @@ PSY.set_r!(br, 2.0)
     @test res["flow_results"].P_from_to[4] == 0.0
     @test res["flow_results"].P_to_from[4] == 0.0
 
-    sys = PSB.build_system(PSB.PSSETestSystems, "psse_240_case_renewable_sys")
-    @test run_powerflow!(sys)
+    # Investigate why this test is failing
```
rodrigomha, line 89: We should open an issue for this

rodrigomha (approved):

PR #19, opened 2023-03-31
Implement dc powerflow

alefcastelli:
This PR to present new methods based on the recent advancements in PowerNetworkMatrices (v0.5). The methods can store data and solve the power flow problem either by using the full PTDF matrix, or the ABA and BA matrices (using LU factorization of ABA).


src/PowerFlowData.jl
```diff
@@ -0,0 +1,106 @@
...
+        bus_type,
+        bus_magnitude,
+        bus_angle,
+        branch_flow_values,
+        network_matrix,
+    )
+end
+
+# version with full PTDF
+function PowerFlowData(
```
jd-lara, line 56: Somewhere around here we should have a connectivity check

test/case_ACTIVSg70k.m
jd-lara, line 1: remove testing files

PR #20, opened 2023-07-03
Multi-period and parallel PowerFlows

alefcastelli:
Multi-period implementation of PowerFlows. Parallel function were tested but not proposed. Might be in next commits if computational bottleneck is solved.


src/PowerFlowData.jl
```diff
@@ -1,19 +1,24 @@
...
+
+struct PowerFlowData{M <: PNM.PowerNetworkMatrix, N, S <: Union{String, Char}}
     bus_lookup::Dict{Int, Int}
     branch_lookup::Dict{String, Int}
-    bus_activepower_injection::Vector{Float64}
-    bus_reactivepower_injection::Vector{Float64}
-    bus_activepower_withdrawals::Vector{Float64}
-    bus_reactivepower_withdrawals::Vector{Float64}
+    bus_activepower_injection::Union{Vector{Float64}, Matrix{Float64}}
+    bus_reactivepower_injection::Union{Vector{Float64}, Matrix{Float64}}
```
jd-lara, line 8: There is no need for this union. This can be just Matrix

src/PowerFlowData.jl
```diff
@@ -1,20 +1,111 @@
...
+    if timesteps != 0
+        if length(timestep_names) == 0
+            timestep_names = [string(i) for i in 1:timesteps]
+        elseif length(timestep_names) != timesteps
+            error("timestep_names field must have same length as timesteps")
+        end
+    end
+
+    # get data for calculations
     power_network_matrix = PNM.Ybus(sys)
```
jd-lara, line 109: We should add connectivity checks here

src/PowerFlowData.jl
```diff
@@ -40,44 +131,114 @@ function PowerFlowData(::ACPowerFlow, sys::PSY.System)
         bus = PSY.get_component(PSY.Bus, sys, bus_name)
         bus_type[ix] = PSY.get_bustype(bus)
         bus_angles[ix] = PSY.get_angle(bus)
+        # ! shouldn't it be like this?
```
jd-lara, line 134: Yes

src/PowerFlowData.jl
```diff
@@ -40,44 +131,114 @@ function PowerFlowData(::ACPowerFlow, sys::PSY.System)
...
         bus_lookup,
         sys,
     )
 
+    # initialize data
+    init_1 = zeros(n_buses, timesteps)
+    init_2 = zeros(n_branches, timesteps)
+
+    # define fields as matrices whose number of columns is eqault to the number of timesteps
+    bus_activepower_injection_1 = deepcopy(init_1)
```
jd-lara, line 166: using 

```suggestion
    bus_activepower_injection_1 = similar(init_1)
```

might be faster than copies

src/PowerFlowData.jl
```diff
@@ -40,44 +131,114 @@ function PowerFlowData(::ACPowerFlow, sys::PSY.System)
...
+    bus_reactivepower_injection_1 = deepcopy(init_1)
+    bus_activepower_withdrawals_1 = deepcopy(init_1)
+    bus_reactivepower_withdrawals_1 = deepcopy(init_1)
+    bus_magnitude_1 = deepcopy(init_1)
+    bus_angles_1 = deepcopy(init_1)
+    branch_flow_values_1 = deepcopy(init_2)
+
+    # initial values related to first timestep allocated in the first column
+    bus_activepower_injection_1[:, 1] .= bus_activepower_injection
```
jd-lara, line 175: why the `_1` nomenclature?

src/common.jl
```diff
@@ -63,69 +63,92 @@ end
...
-)
-    y[:] .= transpose(A) * x
-    return
-end
-
-# virtual case: all lines
-
-function my_mul_mt!(
-    y::Vector{Float64},
+# # sparse case (ABA and BA)
```
jd-lara, line 66: remove this code if it isn't going to be used.

src/post_processing.jl
```diff
@@ -535,6 +521,90 @@ function write_results(
...
+    # get branches from/to buses
+    from_bus = Vector{Int}(undef, length(branches))
+    to_bus = Vector{Int}(undef, length(branches))
+    for (i, branch) in enumerate(branches)
+        br = PSY.get_component(PSY.ACBranch, sys, branch)
+        from_bus[i] = PSY.get_number(PSY.get_arc(br).from)
+        to_bus[i] = PSY.get_number(PSY.get_arc(br).to)
+    end
+
+    # if length(data.timestep_map) == 1
```
jd-lara, line 555: why is this commented out?

src/post_processing.jl
```diff
@@ -535,6 +521,90 @@ function write_results(
...
+    # end
+    # return
+end
+
+# TODO: multi-period still to implement
+"""
+Returns a dictionary containing the AC power flow results.
+
+WARNING: single-period evalution only is suppoerted at the moment. Resulting
+dictionary will therefore feature just one key linked to one DataFrame.
```
jd-lara, line 598:
```suggestion
Only single-period evaluation is supported at the moment for AC Power flows. Resulting
dictionary will therefore feature just one key linked to one DataFrame.
```

src/solve_dc_powerflow.jl
```diff
@@ -114,3 +201,70 @@ function solve_powerflow(
...
+Each DataFrame containts the flows and angles.
+
+# Arguments:
+- `data::PTDFPowerFlowData`:
+        PowerFlowData structure containing the system's data per each timestep
+        considered, as well as the PTDF matrix.
+- `sys::PSY.System`:
+        container gathering the system data.
+"""
+function solve_powerflow(
```
jd-lara, line 220:
```suggestion
function solve_powerflow!(
```

src/solve_dc_powerflow.jl
```diff
@@ -114,3 +201,70 @@ function solve_powerflow(
...
+    data::PTDFPowerFlowData,
+    sys::PSY.System;
+)
+    solve_powerflow!(data)
+    return write_results(data, sys)
```
jd-lara, line 225:
```suggestion
    return write_results!(data, sys)
```

src/solve_dc_powerflow.jl
```diff
@@ -114,3 +201,70 @@ function solve_powerflow(
...
+Each DataFrame containts the flows and angles.
+
+# Arguments:
+- `data::ABAPowerFlowData`:
+        PowerFlowData structure containing the system's data per each timestep
+        considered, as well as the ABA and BA matrices.
+- `sys::PSY.System`:
+        container gathering the system data.
+"""
+function solve_powerflow(
```
jd-lara, line 242:
```suggestion
function solve_powerflow!(
```

test/test/Project.toml
jd-lara, line 1: What is this file about?

test/test_powerflow_data.jl
```diff
@@ -5,3 +5,12 @@
     PowerFlowData(PTDFDCPowerFlow(), sys)
     PowerFlowData(vPTDFDCPowerFlow(), sys)
 end
+
+@testset "PowerFlowData multiperiod" begin
+    sys = PSB.build_system(PSB.PSITestSystems, "c_sys14"; add_forecasts = false)
+    timesteps = 24
+    @warn "multiperiod AC still to implement"
```
jd-lara, line 12: we don't need to log a warning for this

```suggestion
# TODO: "multiperiod AC still to implement"
```

jd-lara (changes requested):
The PR still needs some improvements

src/PowerFlowData.jl
alefcastelli: solved

src/PowerFlowData.jl
alefcastelli: done

src/PowerFlowData.jl
alefcastelli: to differentiate with respect to the vector of initial values (e.g., bus_activepower_injection and bus_activepower_injection_1)

src/post_processing.jl
alefcastelli: I forgot the remove that. Now it is done.

src/PowerFlowData.jl
alefcastelli: fixed

src/common.jl
alefcastelli: done

test/test/Project.toml
alefcastelli: removed

src/solve_dc_powerflow.jl
alefcastelli: I have defined two function, one that overwrites the input (solve_powerflows!), and one that returns a dictionary of data frames (solve_powerflows). It is true though that "solve_powerflows" calls "solve_powerflows!".

shouldn't be better to name the function writing a dictionary of dataframes with no "!" ?

src/solve_dc_powerflow.jl
alefcastelli: same as above

src/solve_dc_powerflow.jl
alefcastelli: same as above

src/PowerFlowData.jl
alefcastelli: This is how I found it. Do you want to add the connectivity check as for the DC case with the ABA and BA matrix? Or do you refer to something different?

PR #26, opened 2023-10-30
Jd/ac power flow refactor

jd-lara:
WIP refactoring the AC PowerFlow to support reactive power checks.


rodrigomha (approved):
This is great.

Remaining aspects is a test for the 14bus system with enforcing q limits, and updating the system if a bus change to PQ (from PV) if a Q limit is hit.

Once the tests are ready this looks good to merge.

PR #42, opened 2024-07-15
PSSE Exporter Part 1: Add Tests and Formalize Exporter Implementation

GabrielKS:
This PR picks up the effort to implement a Sienna to PSS/E exporter, an effort started by @HaleyRoss [here](https://github.com/NREL-Sienna/PowerFlows.jl/pull/32) (see also my cleanup of that code to work with PSY4 [here](https://github.com/NREL-Sienna/PowerFlows.jl/commit/d5e054ae6cc65c8cf217aaffe0eea607f4db92af)). Very much a draft for now, but I'll be building out a public interface to perform this export in the next few weeks.


GabrielKS:
Here's my Git plan:

- `hrgks/psse_exporter_psy4` is a "dev" branch
- For each major part of this effort (currently we have identified three), I'll submit a PR into there
- We'll avoid merging `hrgks/psse_exporter_psy4` into `main` until I can be more certain that there will be no more breaking interface changes (probably not until the end of part two)

This lets me submit PRs for code review along the way, rather than having one massive code review at the end, while also letting the interface evolve as I discover more requirements of it.

GabrielKS:
The current exporter interface is:
```julia
# Initialize an exporter from a Sienna system
PSSEExporter(PSY.System, psse_version::Symbol)

# Whenever you have new data to export, update the exporter with either a `PowerFlowData` or
# a modified `System` (which must be fundamentally the same system, just with some values changed):
update_exporter!(exporter::PSSEExporter, data::PowerFlowData)  # preferred
update_exporter!(exporter::PSSEExporter, data::System)  # less preferred

# Actually perform an export
write_export(exporter::PSSEExporter, scenario_name::AbstractString, year::Int, export_location::AbstractString)
```
and I expect this will probably stay roughly the same throughout Part 1 — though I may change up the arguments a bit — but will probably change substantially in Part 2.

GabrielKS:
The current status of Part 1 is that the above interface is fully implemented and round-trip tested up to a certain level, but the implementation still relies on old code. I'll now work on a from-scratch reimplementation of the exporter.

src/psse_export.jl
```diff
@@ -0,0 +1,599 @@
...
+
+"""
+Update the `PSSEExporter` with new `data`.
+
+# Arguments:
+  - `exporter::PSSEExporter`: the exporter to update
+  - `data::PSY.PowerFlowData`: the new data. Must correspond to the `System` with which the
+    exporter was constructor
+"""
+function update_exporter!(exporter::PSSEExporter, data::PowerFlowData)
```
jd-lara, line 73: For reference since we discussed it already. Relying on deep copies can be problematic for systems with large time series data sets.

src/psse_export.jl
```diff
@@ -0,0 +1,599 @@
...
+  - `data::PSY.System`: system containing the new data. Must be fundamentally the same
+  `System` as the one with which the exporter was constructed, just with different values
+"""
+function update_exporter!(exporter::PSSEExporter, data::PSY.System)
+    _validate_same_system(exporter.system, data) || throw(
+        ArgumentError(
+            "System passed to update_exporter must be the same system as the one with which the exporter was constructed, just with different values",
+        ),
+    )
+    exporter.system = deepcopy(data)
```
jd-lara, line 93: missing return

src/psse_export.jl
```diff
@@ -0,0 +1,599 @@
...
+    for original_number in buses
+        if original_number in 1:999_997
+            mapping[original_number] = original_number
+            push!(used_numbers, original_number)
+        end
+    end
+    for original_number in buses
+        haskey(mapping, original_number) && continue
+        new_number = original_number
+        new_number %= 1_000_000
```
jd-lara, line 186: we should consider addressing this in the parser too. I understand the safety logic of never creating a bus number that PSSe can recognize but it makes a mess to write back

src/psse_export.jl
```diff
@@ -0,0 +1,599 @@
...
+
+    for branch in branches
+        I = md["bus_number_mapping"][PSY.get_number(PSY.get_from_bus(branch))]
+        J = md["bus_number_mapping"][PSY.get_number(PSY.get_to_bus(branch))]
+        CKT = branch_name_mapping[((I, J), PSY.get_name(branch))]
+        @assert !(first(CKT) in ['&', '@', '*'])  # Characters with a special meaning in this context
+        CKT = _psse_quote_string(CKT)
+        R = PSY.get_r(branch)
+        X = PSY.get_x(branch)
+        B = 0.0  # TODO iron out the details of B vs. BI, BJ
```
jd-lara, line 519: The summary about this issue is that the line model in PSSe uses a data model with 2 shunt elements per side and this creates a mess in terms of writing back the data since PSY supports only one imaginary shunt per side. 


image

src/psse_exporter/psse_exporter.jl
```diff
@@ -1,3 +1,47 @@
...
+    )::String
+        sid = device_dict["source_id"]
+        key = join((length(sid) == 6) ? [sid[2], sid[3], sid[5]] : last(sid, 3), "_")
+        return branch_map[key]
+    end
+    return md_bus_name_formatter
+end
+
 # TODO document this function
 function Write_Sienna2PSSE(sys::System, scenario_name::String, year::Int64;
```
jd-lara, line 46:
```suggestion
function write_Sienna2PSSE(sys::System, scenario_name::String, year::Int64;
```

GabrielKS: Deleted entirely

src/psse_export.jl
GabrielKS: added

src/psse_export.jl
GabrielKS: Yes — will change this in Part 2

GabrielKS:
@jarry7 verifies that PSS/E can read the export as currently written under certain modes but not others. I'm going to merge this into the dev branch now to continue with Part 2 and readdress this later.

Issue #43, opened 2024-07-22
Egregious Code Duplication in PowerFlows


GabrielKS:
The block:
```julia
    for (bus_no, ix) in bus_lookup
        bus_name = temp_bus_map[bus_no]
        bus = PSY.get_component(PSY.Bus, sys, bus_name)
        bus_type[ix] = PSY.get_bustype(bus)
        if bus_type[ix] == PSY.ACBusTypes.REF
            bus_angles[ix] = 0.0
        else
            bus_angles[ix] = PSY.get_angle(bus)
        end
        bus_magnitude[ix] = PSY.get_magnitude(bus)
    end
```

appears four times in `PowerFlowData.jl`. I'm sure there is more of this.

jd-lara:
This might have been me.

rodrigomha:
Closed via #45

Issue #41, opened 2024-07-15
`P_from_to` and `P_to_from` are erroneously zero in DC power flow


GabrielKS:
In a power flow output, `P_from_to` is the active power flowing through the line at its `from` bus and `P_to_from` is the active power flowing through the line at its `to` bus. These are not necessarily the same magnitude due to line losses, but they should at least be comparable in magnitude. In the results of an AC power flow, this holds:
<img width="938" alt="AC power flow screenshot (correct behavior)" src="https://github.com/user-attachments/assets/70525978-4c1b-4e31-92fd-4d2af7716e7d">
but in the results of a DC power flow, one of the `P` values for each row seems to be zero:
<img width="929" alt="DC power flow screenshot (incorrect behavior)" src="https://github.com/user-attachments/assets/74b4b47d-fe8f-432f-a827-b40fd69b72f4">

rodrigomha:
I'm confused about this implementation. Here is the code that writes this in the DataFrame: https://github.com/NREL-Sienna/PowerFlows.jl/blob/558a3374d901c984c82e4c9d4d8c15be801f9190/src/post_processing.jl#L491

```
# get flows on each line
    P_from_to_vect = zeros(length(branches))
    Q_from_to_vect = zeros(length(branches))
    P_to_from_vect = zeros(length(branches))
    Q_to_from_vect = zeros(length(branches))
    for i in 1:length(branches)
        if branch_flow_values[i] >= 0
            P_from_to_vect[i] = branch_flow_values[i]
            P_to_from_vect[i] = 0
        else
            P_from_to_vect[i] = 0
            P_to_from_vect[i] = branch_flow_values[i]
        end
    end
```

Based on this implementation, if the solution is positive for the branch, it assigns the proper value to the from_to (probably correct). However if the solution is negative, it's assigned the same value to the `to_from` (probably incorrect). I'm guessing that in the solution of the DC PowerFlow, the solution is always for `from_to`, so in this case what it needs to happen is the following:
```
# get flows on each line
    P_from_to_vect = zeros(length(branches))
    Q_from_to_vect = zeros(length(branches))
    P_to_from_vect = zeros(length(branches))
    Q_to_from_vect = zeros(length(branches))
    for i in 1:length(branches)
        if branch_flow_values[i] >= 0
            P_from_to_vect[i] = branch_flow_values[i]
            P_to_from_vect[i] = 0
        else
            P_from_to_vect[i] = 0
[...]

Issue #40, opened 2024-07-15
Don't say OPF when we mean PF


GabrielKS:
A trivial but important "bug": sometimes in the documentation, we incorrectly refer to a power flow problem as an _optimal_ power flow problem. As someone new to the field, this confused me. Example:
https://github.com/NREL-Sienna/PowerFlows.jl/blob/558a3374d901c984c82e4c9d4d8c15be801f9190/src/PowerFlowData.jl#L98

rodrigomha:
Closed via #45

PR #45, opened 2024-08-06
Fix Bugs/Issues of PowerFlow

### LINKED ISSUE #44: Bug in the weeds of `_reactive_power_redistribution_pv`
The helper function `_reactive_power_redistribution_pv`:
https://github.com/NREL-Sienna/PowerFlows.jl/blob/558a3374d901c984c82e4c9d4d8c15be801f9190/src/post_processing.jl#L283
takes a system, a reactive power, and a bus; and distributes that reactive power among the devices that contribute to the bus. We would therefore expect that after this function is run, the sum of the reactive power among the relevant devices is equal to the `Q_gen` provided to the function. However:

Screenshot 2024-07-22 at 3 44 17 PM

it isn't.

rodrigomha:
Closes #44 #43 #40 #41


rodrigomha:
@GabrielKS @jd-lara The bug in the Q redistribution was that the units was hitting limits in iterative process of the remaining Q, but it was not correctly substracted to the residual in that case. This was fixed by replacing `reallocated_q += q_frac` by  `reallocated_q += q_set_point - current_q` and properly checking the condition of the limits by using `q_frac + current_q` rather than the clamped value.

rodrigomha:
Regarding the tests: Tests are failing for the net_flows with the update of #41.

@GabrielKS @jd-lara Let me know if we are OK with adding the symmetrical and I will update the tests to change the zeros for negative values if needed, or going back to only having one direction and the other zero.

GabrielKS:
> Regarding the tests: Tests are failing for the net_flows with the update of #41.
> 
> @GabrielKS @jd-lara Let me know if we are OK with adding the symmetrical and I will update the tests to change the zeros for negative values if needed, or going back to only having one direction and the other zero.

I like the behavior you have implemented.

GabrielKS (changes requested):
Looks good aside from the test failures.

test/test_dc_powerflow.jl
```diff
@@ -28,8 +28,7 @@
     # CASE 1: ABA and BA matrices
     solved_data_ABA = solve_powerflow(DCPowerFlow(), sys)
     @test isapprox(
-        solved_data_ABA["1"]["flow_results"].P_from_to +
-        solved_data_ABA["1"]["flow_results"].P_to_from,
+        solved_data_ABA["1"]["flow_results"].P_from_to,
```
GabrielKS, line 31: Shouldn't we also test that the value of `P_to_from` is the opposite of `P_from_to`?

src/post_processing.jl
```diff
@@ -405,12 +405,14 @@ function _reactive_power_redistribution_pv(sys::PSY.System, Q_gen::Float64, bus:
                 break
             end
             it += 1
-            if it > 1
+            if it > 5
```
GabrielKS, line 408: Maybe define this as a constant somewhere to reduce magic numbers?

GabrielKS (changes requested):

rodrigomha:
@GabrielKS It is ready for review.

There is now one kwarg when calling `solve_ac_powerflow!`: `max_redistribution_iterations` that controls the max number of redistribution iterations

GabrielKS (approved):
LGTM

PR #48, opened 2024-08-19
PSSE Exporter Part 2: Integrate with PowerSimulations

GabrielKS:
This PR continues the effort from https://github.com/NREL-Sienna/PowerFlows.jl/pull/42 and https://github.com/NREL-Sienna/PowerFlows.jl/pull/29 (companion to https://github.com/NREL-Sienna/PowerSimulations.jl/pull/1040). Here we seek to create a standardized interface to either run a PF power flow or export a PSS/E file for each iteration of a PSI simulation.

### DIFF
--- a/src/common.jl
+++ b/src/common.jl
@@ -1,10 +1,6 @@
-function get_total_p(l::PSY.PowerLoad)
-    return PSY.get_active_power(l)
-end
-
-function get_total_q(l::PSY.PowerLoad)
-    return PSY.get_reactive_power(l)
-end
+_SingleComponentLoad = Union{PSY.PowerLoad, PSY.ExponentialLoad, PSY.InterruptiblePowerLoad}
+get_total_p(l::_SingleComponentLoad) = PSY.get_active_power(l)
+get_total_q(l::_SingleComponentLoad) = PSY.get_reactive_power(l)
 
 function get_total_p(l::PSY.StandardLoad)
     return PSY.get_constant_active_power(l) +
@@ -16,14 +12,6 @@ function get_total_q(l::PSY.StandardLoad)
     return PSY.get_constant_reactive_power(l) +
            PSY.get_current_reactive_power(l) +
            PSY.get_impedance_reactive_power(l)
-end
-
-function get_total_p(l::PSY.ExponentialLoad)
-    return PSY.get_active_power(l)
-end
-
-function get_total_q(l::PSY.ExponentialLoad)
-    return PSY.get_reactive_power(l)
 end
 
 function _get_injections!(
--- a/src/common.jl
+++ b/src/common.jl
@@ -123,7 +123,6 @@ function make_dc_powerflowdata(
     timestep_names,
     power_network_matrix,
     aux_network_matrix,
-    extra_data,
     n_buses,
     n_branches,
     bus_lookup,
@@ -143,7 +142,6 @@ function make_dc_powerflowdata(
         time_steps,
         power_network_matrix,
         aux_network_matrix,
-        extra_data,
         n_buses,
         n_branches,
         bus_lookup,
@@ -162,7 +160,6 @@ function make_powerflowdata(
     time_steps,
     power_network_matrix,
     aux_network_matrix,
-    extra_data,
     n_buses,
     n_branches,
     bus_lookup,
@@ -247,6 +244,5 @@ function make_powerflowdata(
         power_network_matrix,
         aux_network_matrix,
         neighbors,
-        extra_data,
     )
 end
--- a/src/common.jl
+++ b/src/common.jl
@@ -116,3 +116,134 @@ function my_mul_mt(
     end
     return y
 end
+
+function make_dc_powerflowdata(
+    sys,
+    time_steps,
+    timestep_names,
+    power_network_matrix,
+    aux_network_matrix,
+    n_buses,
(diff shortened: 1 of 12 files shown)


GabrielKS:
Tests depend on https://github.com/NREL-Sienna/InfrastructureSystems.jl/pull/408

src/common.jl
```diff
@@ -116,3 +116,133 @@ function my_mul_mt(
...
+    bus_lookup,
+    branch_lookup,
+    temp_bus_map,
+    branch_types,
+    bus_reactivepower_bounds,
+    timestep_map,
+    valid_ix,
+    neighbors,
+)
+    # TODO: bus_type might need to also be a Matrix since the type can change for a particular scenario
```
jd-lara, line 174: Lets do this change right away

src/psse_export.jl
```diff
@@ -59,6 +61,28 @@ function with_units(f::Function, sys::System, units::Union{PSY.UnitSystem, Strin
     end
 end
 
+# TODO a bit hacky, there may be a better way (round trip JSON serialization is too slow).
```
jd-lara, line 64: @daniel-thom thoughts on this?

src/psse_export.jl
```diff
@@ -1020,3 +1073,15 @@ function PSY.System(raw_path::AbstractString, md::Dict)
     # TODO remap everything else! Should be reading all the keys in `md`
     return sys
 end
+
+# TODO handle kwargs
+make_power_flow_container(pfem::PSSEExportPowerFlow, sys::PSY.System; kwargs...) =
```
jd-lara, line 1078: Are you planning this into the public API?

src/solve_dc_powerflow.jl
```diff
@@ -12,7 +12,7 @@ const PTDFPowerFlowData = PowerFlowData{
 }
 
 # ? change this to have a more detailed definition ?
-const vPTDFPowerFlowData = PowerFlowData{}
+const vPTDFPowerFlowData = PowerFlowData{<:Any, <:Any}
```
jd-lara, line 15: we want to figure this out before merging and releasing

src/psse_export.jl
GabrielKS: If we're still planning to not expose `PowerFlowData` to the public then no, since the other methods of this return a `PowerFlowData`. But I've [been thinking](https://github.com/NREL-Sienna/PowerFlows.jl/issues/38#issuecomment-2442183095) that we might need like a "private/internal but stable" API as well as a public API — I need a stable function like this to be able to call from PowerSimulations.

src/solve_dc_powerflow.jl
GabrielKS: yeah this is https://github.com/NREL-Sienna/PowerFlows.jl/issues/50

src/psse_export.jl
daniel-thom: If people need this use case, I'm fine with it. Copying large numbers of time series arrays can be slow. Does it work, though? I'm wondering if the copied components have a reference to the original TimeSeriesManager. This code runs when you add a component to the system.
```
    refs = SharedSystemReferences(;
        time_series_manager = data.time_series_manager,
        supplemental_attribute_manager = data.supplemental_attribute_manager,
    )
    set_shared_system_references!(component, refs)
```
The data will not be copied. But the component would have a reference to another system's time series manager.

daniel-thom: Some options:
1. Remove the shared references between system and components. It causes subtle bugs like this. Benefits are not worth the risks. The Python version does not have them.
2. Reorganize the shared references object. `SystemData` could have a field called `shared_system_references` which contains `time_series_manager` and `supplemental_attribute_manager`. When each component is added, it gets assigned a reference to that field. Then the time_series_manager is hot-swappable.
3. Keep the existing code but add a loop over all components and reassign `time_series_manager` and `supplemental_attribute_manager`.

src/common.jl
GabrielKS: In private discussion with @jd-lara, the resolution to this (old) TODO is to wait until we implement more multi-period AC power flow infrastructure.

src/psse_export.jl
GabrielKS: Strictly speaking, for my use case it should not be a problem that the resulting system's components have references to another system's time series manager, because systems I copy this way ought never be exposed to the user and my private use of them does not involve touching that stuff. But it's a sufficiently weird result that I'm willing to change it anyway. I don't really have the time/appetite to go mucking about with how we handle shared references right now (and it seems like an implementation detail we could easily change later anyway), so I'll go with option 2.

src/solve_dc_powerflow.jl
GabrielKS: Done, see https://github.com/NREL-Sienna/PowerFlows.jl/issues/50#issuecomment-2458075639

src/psse_export.jl
GabrielKS: Done

src/common.jl
GabrielKS: Logged at https://github.com/NREL-Sienna/PowerFlows.jl/issues/52

GabrielKS:
Merging this to proceed with a review of all the recent changes ensemble.

PR #54, opened 2024-12-06
Feature/nr_pf_solver

rbolgaryn:
Integrate the implementation of the Newton-Raphson Power Flow calculation leveraging KLU factorization


Project.toml
```diff
@@ -22,6 +23,7 @@ DataStructures = "0.18"
 Dates = "1"
 InfrastructureSystems = "2"
 JSON3 = "1"
+KLU = "0.6.0"
```
GabrielKS, line 26: Can you specify this version more flexibly? See e.g. https://github.com/NREL-Sienna/PowerNetworkMatrices.jl/blob/0239dfcd140bc60415c2e0abdf3105dab13e3af7/Project.toml#L23

src/PowerFlows.jl
```diff
@@ -4,7 +4,8 @@ export solve_powerflow
 export solve_ac_powerflow!
 export PowerFlowData
 export DCPowerFlow
-export ACPowerFlow
```
GabrielKS, line 7: For backwards compatibility, we may want something like
```julia
const ACPowerFlow = NLSolveACPowerFlow
```
somewhere.

src/nlsolve_ac_powerflow.jl
```diff
@@ -30,29 +30,33 @@ solve_ac_powerflow!(sys)
 solve_ac_powerflow!(sys, method=:newton)
 ```
 """
-function solve_ac_powerflow!(system::PSY.System; kwargs...)
```
GabrielKS, line 33: Could also do a backwards compatibility fix here

src/nlsolve_ac_powerflow.jl
```diff
@@ -30,29 +30,33 @@ solve_ac_powerflow!(sys)
...
+    kwargs...,
+)
     #Save per-unit flag
     settings_unit_cache = deepcopy(system.units_settings.unit_system)
     #Work in System per unit
     PSY.set_units_base_system!(system, "SYSTEM_BASE")
     check_reactive_power_limits = get(kwargs, :check_reactive_power_limits, false)
     data = PowerFlowData(
-        ACPowerFlow(; check_reactive_power_limits = check_reactive_power_limits),
+        NLSolveACPowerFlow(; check_reactive_power_limits = check_reactive_power_limits),
```
GabrielKS, line 44: Now that you have a reference to `pf`, can't you use that here?

src/nlsolve_ac_powerflow.jl
```diff
@@ -30,29 +30,33 @@ solve_ac_powerflow!(sys)
...
     max_iterations = DEFAULT_MAX_REDISTRIBUTION_ITERATIONS
-    res = _solve_powerflow!(data, check_reactive_power_limits; kwargs...)
-    if res.f_converged
-        write_powerflow_solution!(system, res.zero, max_iterations)
+    converged, x = _solve_powerflow!(pf, data, check_reactive_power_limits; kwargs...)
+    if converged
+        write_powerflow_solution!(system, x, max_iterations)
         @info("PowerFlow solve converged, the results have been stored in the system")
         #Restore original per unit base
         PSY.set_units_base_system!(system, settings_unit_cache)
```
GabrielKS, line 54: No need to do this now but just flagging that this is a good use case for https://github.com/NREL-Sienna/PowerFlows.jl/blob/465e92fcf921c460f1bd9e0c042c2cc335897c2f/src/psse_export.jl#L47 (and a good argument for moving it someplace more central)

src/nlsolve_ac_powerflow.jl
```diff
@@ -30,29 +30,33 @@ solve_ac_powerflow!(sys)
...
-        return res.f_converged
+        return converged
     end
-    @error("The powerflow solver returned convergence = $(res.f_converged)")
+    @error("The powerflow solver returned convergence = $(converged)")
     PSY.set_units_base_system!(system, settings_unit_cache)
```
GabrielKS, line 58: duplicate?

src/nlsolve_ac_powerflow.jl
```diff
@@ -124,27 +128,32 @@ function _check_q_limit_bounds!(data::ACPowerFlowData, zero::Vector{Float64})
...
             end
         end
     else
-        return _nlsolve_powerflow(data; nlsolve_kwargs...)
+        return _nlsolve_powerflow(pf, data; nlsolve_kwargs...)
     end
 end
 
-function _nlsolve_powerflow(data::ACPowerFlowData; nlsolve_kwargs...)
+function _nlsolve_powerflow(
```
GabrielKS, line 152: Let's rename this now that one of the methods doesn't use NLsolve

src/nlsolve_ac_powerflow.jl
```diff
@@ -153,5 +162,100 @@ function _nlsolve_powerflow(data::ACPowerFlowData; nlsolve_kwargs...)
...
-    return res
+    return res.f_converged, res.zero
+end
+
+function _nlsolve_powerflow(pf::KLUACPowerFlow, data::ACPowerFlowData; nlsolve_kwargs...)
+    pf = PolarPowerFlow(data)
+    #J_function = PowerFlows.PolarPowerFlowJacobian(data, pf.x0)
+
+    maxIter = 30  # TODO
+    tol = 1e-6    # TODO
```
GabrielKS, line 173: It would be good to let these be configurable via the kwargs

src/nlsolve_ac_powerflow.jl
```diff
@@ -153,5 +162,100 @@ function _nlsolve_powerflow(data::ACPowerFlowData; nlsolve_kwargs...)
...
+    i = 0
+
+    Vm = data.bus_magnitude[:]
+    Va = data.bus_angles[:]
+    V = Vm .* exp.(1im * Va)
+
+    # Find indices for each bus type
+    ref = findall(x -> x == PowerSystems.ACBusTypesModule.ACBusTypes.REF, data.bus_type)
+    pv = findall(x -> x == PowerSystems.ACBusTypesModule.ACBusTypes.PV, data.bus_type)
+    pq = findall(x -> x == PowerSystems.ACBusTypesModule.ACBusTypes.PQ, data.bus_type)
```
GabrielKS, line 183: Probably no significant difference, but note that these three lines could be combined to only iterate over the bus types once

src/nlsolve_ac_powerflow.jl
```diff
@@ -153,5 +162,100 @@ function _nlsolve_powerflow(data::ACPowerFlowData; nlsolve_kwargs...)
...
+        j12 = real(dSbus_dVm[[pv; pq], pq])
+        j21 = imag(dSbus_dVa[pq, [pv; pq]])
+        j22 = imag(dSbus_dVm[pq, pq])
+        J = [j11 j12; j21 j22]
+
+        factor_J = KLU.klu(J)
+        dx = -(factor_J \ F)
+
+        #J_function(J_function.Jv, x)
+        #dx = - (J_function.Jv \ F)
```
GabrielKS, line 218: What are these lines?

src/nlsolve_ac_powerflow.jl
```diff
@@ -153,5 +162,100 @@ function _nlsolve_powerflow(data::ACPowerFlowData; nlsolve_kwargs...)
...
+
+    if !converged
+        @error("The powerflow solver with KLU did not converge after $i iterations")
+    else
+        @debug("The powerflow solver with KLU converged after $i iterations")
+    end
+
+    # mock the expected x format, where the values depend on the type of the bus:
+    n_buses = length(data.bus_type)
+    x = Float64[0.0 for _ in 1:(2 * n_buses)]
```
GabrielKS, line 242: Use `repeat`

GabrielKS (changes requested):
A few initial comments. I'll dig into the math more later. Thanks!

src/nlsolve_ac_powerflow.jl
GabrielKS: (and put the defaults into `const`s somewhere)

src/nlsolve_ac_powerflow.jl
```diff
@@ -153,5 +162,100 @@ function _nlsolve_powerflow(data::ACPowerFlowData; nlsolve_kwargs...)
...
+        dSbus_dVa = 1im * diagV * conj(diagIbus - Ybus * diagV)
+
+        j11 = real(dSbus_dVa[[pv; pq], [pv; pq]])
+        j12 = real(dSbus_dVm[[pv; pq], pq])
+        j21 = imag(dSbus_dVa[pq, [pv; pq]])
+        j22 = imag(dSbus_dVm[pq, pq])
+        J = [j11 j12; j21 j22]
+
+        factor_J = KLU.klu(J)
+        dx = -(factor_J \ F)
```
GabrielKS, line 215: Comment from José: do this and previous line in place using functor approach

src/powerflow_types.jl
```diff
@@ -1,6 +1,10 @@
 abstract type PowerFlowEvaluationModel end
 
-Base.@kwdef struct ACPowerFlow <: PowerFlowEvaluationModel
+Base.@kwdef struct NLSolveACPowerFlow <: PowerFlowEvaluationModel
+    check_reactive_power_limits::Bool = false
+end
+
+Base.@kwdef struct KLUACPowerFlow <: PowerFlowEvaluationModel
```
GabrielKS, line 7: Comment from José: go back to one struct for all AC power flow, specify the algorithm with struct parameter or field or something

src/nlsolve_ac_powerflow.jl
rbolgaryn: Added the constants to definitions.jl

src/newton_ac_powerflow.jl
```diff
@@ -0,0 +1,266 @@
...
+        diagIbus = LinearAlgebra.Diagonal(Ybus * V)
+        diagVnorm = LinearAlgebra.Diagonal(V ./ abs.(V))
+        dSbus_dVm = diagV * conj(Ybus * diagVnorm) + conj(diagIbus) * diagVnorm
+        dSbus_dVa = 1im * diagV * conj(diagIbus - Ybus * diagV)
+
+        j11 = real(dSbus_dVa[[pv; pq], [pv; pq]])
+        j12 = real(dSbus_dVm[[pv; pq], pq])
+        j21 = imag(dSbus_dVa[pq, [pv; pq]])
+        j22 = imag(dSbus_dVm[pq, pq])
+        J = [j11 j12; j21 j22]
```
jd-lara, line 220: Leave note that we shouldn't allocate J every time

src/newton_ac_powerflow.jl
```diff
@@ -0,0 +1,266 @@
...
+        Va[pq] .+= dx[(npv + 1):(npv + npq)]
+        Vm[pq] .+= dx[(npv + npq + 1):(npv + 2 * npq)]
+
+        V = Vm .* exp.(1im * Va)
+
+        Vm = abs.(V)
+        Va = angle.(V)
+
+        mis = V .* conj(Ybus * V) - Sbus
+        F = [real(mis[[pv; pq]]); imag(mis[pq])]
```
jd-lara, line 235: should be changed to not allocating

src/newton_ac_powerflow.jl
```diff
@@ -0,0 +1,266 @@
...
+
+    converged = (npv + npq) == 0  # if only ref buses present, we do not need to enter the loop
+
+    while i < maxIter && !converged
+        i += 1
+        diagV = LinearAlgebra.Diagonal(V)
+        diagIbus = LinearAlgebra.Diagonal(Ybus * V)
+        diagVnorm = LinearAlgebra.Diagonal(V ./ abs.(V))
+        dSbus_dVm = diagV * conj(Ybus * diagVnorm) + conj(diagIbus) * diagVnorm
+        dSbus_dVa = 1im * diagV * conj(diagIbus - Ybus * diagV)
```
jd-lara, line 214: All of this code should be more allocating

src/newton_ac_powerflow.jl
```diff
@@ -0,0 +1,266 @@
...
+    V = Vm .* exp.(1im * Va)
+
+    Ybus = data.power_network_matrix.data
+
+    Sbus =
+        data.bus_activepower_injection[:] - data.bus_activepower_withdrawals[:] +
+        1im * (data.bus_reactivepower_injection[:] - data.bus_reactivepower_withdrawals[:])
+
+    mis = V .* conj(Ybus * V) - Sbus
+    F = [real(mis[[pv; pq]]); imag(mis[pq])]
```
jd-lara, line 200: This function should be not allocating

src/newton_ac_powerflow.jl
```diff
@@ -0,0 +1,336 @@
...
+    mis = zeros(Complex{Float64}, length(V))
+    mis_pvpq = view(mis, pvpq)
+    mis_pq = view(mis, pq)
+
+    F = zeros(Float64, npvpq + npq)
+    F_real = view(F, 1:npvpq)
+    F_imag = view(F, npvpq + 1:npvpq + npq)
+
+    mis .= V .* conj.(Ybus * V) .- Sbus
+    @. F_real = real(mis_pvpq)  # In-place assignment to the real part, using views
```
jd-lara, line 226: @rbolgaryn I think you are overwriting F_real here. You might want to try `F_real[:] = `

src/newton_ac_powerflow.jl
```diff
@@ -0,0 +1,726 @@
...
+    # sometimes can allocate so we have to use the for loop above
+    r_dSbus_dVa .= real.(dSbus_dVa)
+    r_dSbus_dVm .= real.(dSbus_dVm)
+    i_dSbus_dVa .= imag.(dSbus_dVa)
+    i_dSbus_dVm .= imag.(dSbus_dVm)
+    return
+end
+
+# this function is for testing purposes only
+function _legacy_dSbus_dV(
```
jd-lara, line 372: If this code is for testing, should be commented out before merging.

src/newton_ac_powerflow.jl
```diff
@@ -0,0 +1,726 @@
...
+    diagV = LinearAlgebra.Diagonal(V)
+    diagVnorm = LinearAlgebra.Diagonal(V ./ abs.(V))
+    diagIbus = LinearAlgebra.Diagonal(Ybus * V)
+    dSbus_dVm = diagV * conj.(Ybus * diagVnorm) + conj.(diagIbus) * diagVnorm
+    dSbus_dVa = 1im * diagV * conj.(diagIbus - Ybus * diagV)
+    return dSbus_dVa, dSbus_dVm
+end
+
+# this function is for testing purposes only
+function _legacy_J(
```
jd-lara, line 385: same here

src/newton_ac_powerflow.jl
```diff
@@ -0,0 +1,726 @@
...
+        # solve inplace - the results are written to F, so that we must use F instead of dx for updating V
+        KLU.klu_l_solve(
+            factor_J._symbolic,
+            factor_J._numeric,
+            size(F, 1),
+            size(F, 2),
+            F,
+            rf,
+        )
+
```
jd-lara, line 623: do this operation around a try catch call in case the factorization fails we can have a graceful exit.

src/newton_ac_powerflow.jl
jd-lara, line 1: lets separate NL Solve calls from the klu solver. Makes the PR easier to read

src/newton_ac_powerflow.jl
```diff
@@ -0,0 +1,726 @@
...
+    i_dSbus_dVa = sparse(rows, cols, Float64(0))
+    i_dSbus_dVm = sparse(rows, cols, Float64(0))
+
+    # maybe use this in the future?
+    # pvpq_rows = pvpq_lookup[rows][pvpq_lookup[rows] .!= 0]
+    # pvpq_cols = pvpq_lookup[cols][pvpq_lookup[cols] .!= 0]
+    # pq_rows = pq_lookup[rows][pq_lookup[rows] .!= 0] 
+    # pq_cols = pq_lookup[cols][pq_lookup[cols] .!= 0]
+
+    J_block = sparse(rows, cols, Float64(0), maximum(rows), maximum(cols))
```
jd-lara, line 561: Wrap this in a function like `get_jacobian` so that we can improve the return of this later

src/newton_ac_powerflow.jl
jd-lara: should move them to the test code not in the main src

src/newton_ac_powerflow.jl
rbolgaryn: I suggest that we delete NLSolve at this point. Is it OK?

jd-lara: Not until we have the same performance.

rbolgaryn: OK

src/newton_ac_powerflow.jl
```diff
@@ -0,0 +1,726 @@
...
+        if !converged || !check_reactive_power_limits || _check_q_limit_bounds!(data, x)
+            return converged, x
+        end
+    end
+    # todo: throw error? set converged to false?
+    @error("could not enforce reactive power limits after $MAX_REACTIVE_POWER_ITERATIONS")
+    return converged, x
+end
+
+function _newton_powerflow(
```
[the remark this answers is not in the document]
rbolgaryn: We are relying on multiple dispatch here to go to KLU or NLSolve from solve_powerflow(...)

GabrielKS (changes requested):
Various notes from a software development perspective

src/nlsolve_ac_powerflow.jl
GabrielKS: This can be done now — `with_units` is now `with_units_base` in PSY

GabrielKS:
Merging now so we can have a consolidated dev branch to work off of — @rbolgaryn will continue to address the above feedback in a new pull request.

PR #76, opened 2025-01-11
Fix bug 74: enforcing reactive power limits for AC power flow

GabrielKS:
Fixes https://github.com/NREL-Sienna/PowerFlows.jl/issues/74:

- Filters out kwargs that shouldn't be passed along to NLSolve
- Updates the `System` with changes that aren't captured in the solver results

@rbolgaryn @annacasavant tagging for visibility, don't necessarily need a review from you.


jd-lara (approved):

rodrigomha (approved):
This looks good to me

PR #79, opened 2025-01-16
Implement power flow-specific (re)active power limits proxies

GabrielKS:
Fixes https://github.com/NREL-Sienna/PowerFlows.jl/issues/39 by implementing the functions `get_reactive_power_limits_for_power_flow` and `get_active_power_limits_for_power_flow`. In all normal cases, these just redirect to PSY's `get_reactive_power_limits` and `get_active_power_limits`. In a few special cases, each either detailed at https://github.com/NREL-Sienna/PowerFlows.jl/issues/39#issuecomment-2594101291 or already implemented elsewhere in the codebase, they instead return a proxy value to effect the behavior we want.

### DIFF
--- a/src/common.jl
+++ b/src/common.jl
@@ -14,6 +14,38 @@ function get_total_q(l::PSY.StandardLoad)
            PSY.get_impedance_reactive_power(l)
 end
 
+"""
+Return the reactive power limits that should be used in power flow calculations and PSS/E
+exports. Redirects to `PSY.get_reactive_power_limits` in all but special cases.
+"""
+get_reactive_power_limits_for_power_flow(gen::PSY.Device) =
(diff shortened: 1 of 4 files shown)


jd-lara (approved):

jd-lara:
@GabrielKS does this need a release of PSY?

GabrielKS:
> @GabrielKS does this need a release of PSY?

No, this PR is independent of PSY. There are still those bugs we identified in PSY but these don't call the problematic methods over there.

The test failure is coming from `fast_deepcopy_system`, which will need a PSY release before `hrgks/psse_exporter_psy4` is merged into main, but not before `gks/principled_power_limits` is merged into `hrgks/psse_exporter_psy4`.

PR #82, opened 2025-01-16
add field "converged" to PowerFlowData, return NaN when not converged

rbolgaryn:
* add converged field to PowerFlowdata
* NaN for not converged time steps


GabrielKS:
Merging now to keep the branches from diverging too much. TODO add docstrings and tests.

PR #89, opened 2025-02-05
update project toml

jd-lara:


rbolgaryn (approved):

jd-lara:
I'll merge with the windows test fail but opened #90 since the function that is failing is in the test code. @rbolgaryn please look into that error.

PR #92, opened 2025-02-05
Replace manual unrolls with macro calls (#86)

jd-lara:
* manually unrolled fastprints replaced with macro

* more manual unrolls replaced with macro calls


jd-lara:
this code was already approved. Waiting for the tests to pass and then merge

PR #93, opened 2025-02-05
address failing test on windows

rbolgaryn:
Fixing test for windows

### DIFF
--- a/test/test_utils/legacy_pf.jl
+++ b/test/test_utils/legacy_pf.jl
@@ -7,7 +7,7 @@ struct LUACPowerFlow <: ACPowerFlowSolverType end  # Only for testing, a basic i
 function _legacy_dSbus_dV(
     V::Vector{Complex{Float64}},
     Ybus::SparseMatrixCSC{Complex{Float64}, Int64},
-)
+)::Tuple{SparseMatrixCSC{Complex{Float64}, Int64}, SparseMatrixCSC{Complex{Float64}, Int64}}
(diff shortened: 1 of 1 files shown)


GabrielKS:
What is this for?

jd-lara:
@rbolgaryn this is still failing

rbolgaryn:
Have not been able to figure out why it is failing on Windows. I just commented out the lines that were causing the problem in the "legacy" implementation. Since it is not used other than for testing, it should be OK. This PR should be ready to merge.

GabrielKS (commented):
If this is just for testing and we're sure we don't need the lines that are commented out here, might as well just delete them?

test/test_utils/legacy_pf.jl
```diff
@@ -127,9 +127,9 @@ function _newton_powerflow(
         @error("The powerflow solver with KLU did not converge after $i iterations")
     else
         Sbus_result = V .* conj(Ybus * V)
-        aux_variables.J = J
-        aux_variables.dSbus_dV_ref =
-            [vec(real.(dSbus_dVa[ref, :][:, pvpq])); vec(real.(dSbus_dVm[ref, :][:, pq]))]
+        # aux_variables.J = J
```
jd-lara, line 130: @rbolgaryn do we need to keep this code here?

GabrielKS (approved):
Maybe add a comment about why we aren't calculating these things here. Otherwise looks fine to me.

jd-lara (approved):

PR #94, opened 2025-02-08
Use views of real, imag parts

luke-kiernan:
In `newton_ac_powerflow.jl`, `[r/i]_dSbus_dVa` consistently equaled the real/imaginary part of `dSbus_dVa`. I refactored things to use views instead, eliminating those variables. `_update_submatrix!` remains non-allocating (though it's perhaps a bit less readable than before).

### DIFF
--- a/src/newton_ac_powerflow.jl
+++ b/src/newton_ac_powerflow.jl
@@ -315,8 +315,9 @@ end
 
 function _update_V!(dx::Vector{Float64}, V::Vector{Complex{Float64}}, Vm::Vector{Float64},
     Va::Vector{Float64},
-    pv::Vector{Int64}, pq::Vector{Int64}, dx_Va_pv::Vector{Int64}, dx_Va_pq::Vector{Int64},
-    dx_Vm_pq::Vector{Int64})
+    pv::Vector{Int64}, pq::Vector{Int64}, dx_Va_pv::UnitRange{Int64}, dx_Va_pq::UnitRange{Int64},
+    dx_Vm_pq::UnitRange{Int64})
+
     for (i, j) in zip(pv, dx_Va_pv)
         Va[i] -= dx[j]
     end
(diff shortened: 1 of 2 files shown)


src/newton_ac_powerflow.jl
```diff
@@ -425,55 +422,44 @@ function _update_dSbus_dV!(rows::Vector{Int64}, cols::Vector{Int64},
...
+    B::SparseMatrixCSC{ComplexF64, Int64},
+    real_rows_A::Vector{Int64},
+    imag_rows_A::Vector{Int64},
     cols_A::Vector{Int64},
-    rows_B::Vector{Int64},
+    real_rows_B::Vector{Int64},
+    imag_rows_B::Vector{Int64},
     cols_B::Vector{Int64},
 )
+    real_B = @view reinterpret(Float64, B)[begin:2:end, :]
```
GabrielKS, line 438: You're `reinterpret`ing a `SparseMatrixCSC` as a `Float64` array? That seems unstable — relies on the internal, private representation of `SparseMatrixCSC` never changing. Is there an accessor you can use here?

src/newton_ac_powerflow.jl
```diff
@@ -425,55 +422,44 @@ function _update_dSbus_dV!(rows::Vector{Int64}, cols::Vector{Int64},
...
+    for idj in eachindex(cols_A)
+        for idi in eachindex(real_rows_A)
+            A[real_rows_A[idi], cols_A[idj]] = real_B[real_rows_B[idi], cols_B[idj]]
+        end
+    end
+    imag_B = @view reinterpret(Float64, B)[2:2:end, :]
```
GabrielKS, line 444: (same here)

GabrielKS (changes requested):
Looks mostly good, but seems like a dealbreaker to rely on the internal representation of `SparseMatrixCSC` the way I think you are.

src/newton_ac_powerflow.jl
GabrielKS: Never mind, I misunderstood `reinterpret`

src/newton_ac_powerflow.jl
GabrielKS: (same here)

GabrielKS:
@luke-kiernan says this is now obsolete; closing.

PR #98, opened 2025-02-14
loss factors as matrix in PowerFlowData

rbolgaryn:
* add a kwarg `calc_loss_factors` to ACPowerFlow (default false)
* write loss factors as an attribute of PowerFlowData - 2D array (rows - bus, columns - time steps)
* initialize the loss factors as `nothing` by default
* test the values of loss factors

### DIFF
--- a/src/PowerFlowData.jl
+++ b/src/PowerFlowData.jl
@@ -233,7 +233,11 @@ function PowerFlowData(
     aux_network_matrix = nothing
     converged = fill(false, time_steps)
     # loss factors order matchn the order of buses in the grid model, and is calculated for sall buses including ref buses (0 for ref buses)
-    loss_factors = pf.calc_loss_factors ? Matrix{Float64}(undef, (n_buses, length(timestep_names))) : nothing
+    loss_factors = if pf.calc_loss_factors
+        Matrix{Float64}(undef, (n_buses, length(timestep_names)))
+    else
+        nothing
+    end
 
     return make_powerflowdata(
         sys,
--- a/src/PowerFlowData.jl
+++ b/src/PowerFlowData.jl
@@ -1,17 +1,3 @@
-"""
-    AuxiliaryVariables
-
-A mutable struct to hold auxiliary variables that are produced by power flow calculations, particularly Jacobian matrix.
-
-# Fields
-- `J::Union{SparseMatrixCSC{Float64, Int}, Nothing}`: A sparse matrix representing the Jacobian matrix after a successfully converged power flow calculation. This is the J matrix from the last step of the Newton-Raphson method. If the power flow calculation does not converge, it is set to `nothing`. Defaults to `nothing`.
-- `dSbus_dV_ref::Union{Vector{Float64}, Nothing}`: A vector representing the partial derivatives of the reference bus apparent power injections with respect to the non-reference bus voltages. Defaults to `nothing`.
-"""
-@kwdef mutable struct AuxiliaryVariables
-    J::Union{SparseMatrixCSC{Float64, Int}, Nothing} = nothing
-    dSbus_dV_ref::Union{Vector{Float64}, Nothing} = nothing
-end
-
 abstract type PowerFlowContainer end
 
 """
@@ -117,7 +103,7 @@ struct PowerFlowData{
     aux_network_matrix::N
     neighbors::Vector{Set{Int}}
     converged::Vector{Bool}
-    aux_variables::Vector{AuxiliaryVariables}
+    loss_factors::Union{Matrix{Float64}, Nothing}
 end
 
 get_bus_lookup(pfd::PowerFlowData) = pfd.bus_lookup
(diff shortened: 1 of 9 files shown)


test/test_multiperiod_ac_powerflow.jl
```diff
@@ -17,30 +9,32 @@
...
+    injs = Matrix(injections)
+    withs = Matrix(withdrawals)
 
-    ##############################################################################
+    data.bus_activepower_injection .= deepcopy(injs[:, 1:time_steps])
+    data.bus_activepower_withdrawals .= deepcopy(withs[:, 1:time_steps])
+    return nothing
+end
+
+# work in progress
```
GabrielKS, line 21: resolve or issue before merge

test/test_utils/legacy_pf.jl
```diff
@@ -42,15 +42,14 @@ function _newton_powerflow(
...
-    aux_variables = data.aux_variables[time_step]
 
     Ybus = data.power_network_matrix.data
 
     # Find indices for each bus type
-    ref = findall(
-        x -> x == PowerSystems.ACBusTypesModule.ACBusTypes.REF,
-        data.bus_type[:, time_step],
-    )
+    # ref = findall(
```
GabrielKS, line 49: Can this be removed if it is not necessary?

test/test_multiperiod_ac_powerflow.jl
```diff
@@ -133,3 +99,24 @@ end
         atol = 1e-9,
     )
 end
+
+@testset "test_loss_factors_case_14" begin
+    sys = build_system(PSITestSystems, "c_sys14"; add_forecasts = false)
+
+    pf_klu = ACPowerFlow(KLUACPowerFlow; calc_loss_factors = true)
```
GabrielKS, line 106: Test that `calc_loss_factors = false` does not calculate loss factors

GabrielKS (changes requested):
A few code comments. Can you add a description to the pull request for future reference?

src/PowerFlowData.jl
```diff
@@ -117,7 +103,7 @@ struct PowerFlowData{
     aux_network_matrix::N
     neighbors::Vector{Set{Int}}
     converged::Vector{Bool}
-    aux_variables::Vector{AuxiliaryVariables}
```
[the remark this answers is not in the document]
rbolgaryn: We can add this structure in the future if we end up needing it

src/PowerFlows.jl
```diff
@@ -16,7 +16,8 @@ export PSSEExporter
 export update_exporter!
 export write_export
 export get_psse_export_paths
-export penalty_factors
+export penalty_factors!
+export penalty_factors_brute_force
```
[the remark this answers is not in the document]
rbolgaryn: As we only use the functions internally, no need to export them

src/post_processing.jl
```diff
@@ -730,9 +730,87 @@ function update_system!(sys::PSY.System, data::PowerFlowData; time_step = 1)
     end
 end
 
+"""
+    penalty_factors!(J::SparseMatrixCSC{Float64, Int64}, dSbus_dV_ref::Vector{Float64}, destination::SubArray{Float64})
```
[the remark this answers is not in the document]
rbolgaryn: We use the function internally only, so no need to make it more generalized.

src/PowerFlows.jl
```diff
@@ -16,7 +16,8 @@ export PSSEExporter
 export update_exporter!
 export write_export
 export get_psse_export_paths
-export penalty_factors
+export penalty_factors!
```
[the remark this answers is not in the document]
rbolgaryn: I agree that we don't need to export them

test/test_multiperiod_ac_powerflow.jl
rbolgaryn: #99

rbolgaryn:
> A few code comments. Can you add a description to the pull request for future reference?

Thank you! I have addressed your comments

GabrielKS (commented):
A few more things and then I think this is good from the code quality point of view.

src/post_processing.jl
```diff
@@ -730,9 +730,83 @@ function update_system!(sys::PSY.System, data::PowerFlowData; time_step = 1)
...
+"""
+
+function penalty_factors_brute_force(
+    data::PowerFlowData;
+    step_size::Float64 = 1e-6,
+    kwargs...,
+)
+
+    # we assume that the bus type for ref bus does not change between time steps
+    ref, _, _ = bus_type_idx(data, 1)
```
GabrielKS, line 790: Aren't we doing a lot of extra work here that we're throwing away?

GabrielKS: Could have `bus_type_idx` take a kwarg for what bus types to calculate that defaults to all three of them

rbolgaryn: good point

GabrielKS: (or a positional argument if that's more performant than a keyword argument)

GabrielKS (approved):
New `bus_type_idx` looks good. We should investigate at some point whether it makes sense to do something like https://github.com/NREL-Sienna/PowerSystems.jl/pull/1216 with it (@luke-kiernan has been working on this elsewhere in the PowerFlows codebase), but I'm fine to leave it as future work since this PR doesn't introduce the issue.

PR #105, opened 2025-02-21
loss factors with the new PR implementation

rbolgaryn:
* add ProgressMeter for multiperiod AC PF and brute-force loss factor calculations
* filter init voltage guess
* implement loss factors for HybridPowerFlow
* use constants for tol and xtol
*r enamed some variables, adjusted signature for HybridPowerFlow to match the other functions
* added a test for loss factors with 2000 bus grid and HybridPowerFlow

TODO: figure out how to handle the PSSE roundtrip testsn that are failing because of the voltage guess correction.


luke-kiernan (approved):
Looks good to me!

PR #109, opened 2025-02-26
hotfix for loss factors at ref bus (1 instead of 0)

rbolgaryn:
This PR fixes the wron value of loss factors calculated for REF bus (should be 1 instead of 0)


jd-lara (approved):

GabrielKS (commented):
Looks fine to me

PR #111, opened 2025-02-28
Lin solve cache 2

rbolgaryn:
* Restructuring of code in nlsolve_powerflow.jl
* Renaming of variables for clarity and readability
* Still work in progress


src/ac_power_flow_residual.jl
```diff
@@ -0,0 +1,105 @@
...
+    P_net::Vector{Float64},
+    Q_net::Vector{Float64},
+    data::ACPowerFlowData,
+    time_step::Int64,
+)
+    # update P_net, Q_net, data.bus_angles, data.bus_magnitude based on X
+    Yb = data.power_network_matrix.data
+    bus_types = view(data.bus_type, :, time_step)
+    for (ix, bt) in enumerate(bus_types)
+        if bt == PSY.ACBusTypes.REF
```
jd-lara, line 62: Use functional approach @luke-kiernan implemented for the Jacobian and the Flow

src/nlsolve_powerflow.jl
```diff
@@ -64,84 +82,84 @@ function _dogleg!(p::Vector{Float64},
...
+    X.r_predict .+= X.r
     rho =
-        (sum(abs2, cache.r) - sum(abs2, pf.residual)) /
-        (sum(abs2, cache.r) - sum(abs2, cache.r_predict))
+        (sum(abs2, X.r) - sum(abs2, R.Rv)) /
+        (sum(abs2, X.r) - sum(abs2, X.r_predict))
     if rho > eta
         # Successful iteration
-        cache.r .= pf.residual
+        X.r .= R.Rv
```
jd-lara, line 111: I don't like to have single letter variables. Here I can't know what `X` is. In the developers guidelines for Sienna we discourage this.

src/nlsolve_powerflow.jl
```diff
@@ -64,84 +82,84 @@ function _dogleg!(p::Vector{Float64},
 end
 
 function _trust_region_step(time_step::Int,
-    cache::NLCache,
+    X::StateVectorCache,
     linSolveCache::KLULinSolveCache{Int32},
-    pf::PolarPowerFlow,
+    R::ACPowerFlowResidual,
```
jd-lara, line 87: This change makes the code more confusing

rbolgaryn: renamed to not have single-letter variable names

src/PowerFlowData.jl
```diff
@@ -233,10 +235,10 @@ function PowerFlowData(
     aux_network_matrix = nothing
     converged = fill(false, time_steps)
     # loss factors order matches the order of buses in the grid model, and is calculated for all buses including ref buses (equals 0 for ref buses)
-    loss_factors = if pf.calc_loss_factors
-        Matrix{Float64}(undef, (n_buses, length(timestep_names)))
+    if calc_loss_factors
```
luke-kiernan, line 238: I'd prefer it as it was, `loss_factors = `: I'm actually surprised Julia allows for this, that the `loss_factors` local variable inside the `if` is recognized outside the `if`.

src/PowerFlowData.jl
```diff
@@ -188,7 +189,8 @@ function PowerFlowData(
     sys::PSY.System;
```
luke-kiernan, line 189: Looks like `pf` isn't used in the body of this, so we can change this to `::ACPowerFlow{<:ACPowerFlowSolverType}`

src/PowerFlowData.jl
```diff
@@ -104,6 +104,7 @@ struct PowerFlowData{
     neighbors::Vector{Set{Int}}
     converged::Vector{Bool}
     loss_factors::Union{Matrix{Float64}, Nothing}
+    calc_loss_factors::Bool
```
jd-lara, line 107: use full words in the public interface. 

```suggestion
    calculate_loss_factors::Bool
```

luke-kiernan (changes requested):
Two small changes to `PowerFlowData.jl`

src/PowerFlowData.jl
jd-lara: @rbolgaryn keep the code as it is, single like if statements are discouraged and not used in the Sienna source code

src/PowerFlowData.jl
jd-lara: wait, the `calculate_loss_factors` should be an opt in only. Why have two ?

rbolgaryn: yes, it is needed to suppress the loss factors calculation when the data instance was initialized with calc_loss_factors = true

luke-kiernan (approved):
Looks good to me

src/powerflow_method.jl
```diff
@@ -0,0 +1,412 @@
+"""Abstract supertype for all variations of Newton-Raphson."""
+abstract type AbstractNewtonRaphsonMethod end
+"""Basic, no-frills Newton-Raphson."""
+struct SimpleMethod <: AbstractNewtonRaphsonMethod end
+"""Use iterative refinement to get better accuracy when solving J(x)*Δx = -f(x)."""
+struct RefinementMethod <: AbstractNewtonRaphsonMethod end
```
jd-lara, line 6: I don't think we need two variants of NR. Is there a major value on having separate ones?

luke-kiernan: Not really: doing NR over again from the beginning, but with iterative refinement, seems pretty dumb. I have yet to find a case where iterative refinement works, when straight NR doesn't. What did you have in mind, when you asked me to add iterative refinement as a fallback strategy? I could look into using the condition number of LU factorization to figure out if iterative refinement is worthwhile: KLU includes code for calculating that.

src/PowerFlowData.jl
jd-lara: Again, I've mentioned this before that it isn't a good practice to have options on the source code that are for testing. Mixing up testing options with source code is problematic.

PR #120, opened 2025-03-20
updates to documenter 1.0, added md to formatter

annacasavant:
- updated to documenter ^1.0 
- added markdown to formatter
@kdayday


jd-lara:
we should delay this until we actually finish the methods

jd-lara (approved):

PR #122, opened 2025-03-20
changes for Transformer3W

rbolgaryn:
Changes implemented for compatibility with Transformer3W. This PR depends on:

https://github.com/NREL-Sienna/PowerNetworkMatrices.jl/pull/101

https://github.com/NREL-Sienna/PowerSystems.jl/pull/1306


jd-lara:
I switched this PR to target a PSY5 branch

src/PowerFlowData.jl
```diff
@@ -84,6 +84,7 @@ struct PowerFlowData{
 } <: PowerFlowContainer
     bus_lookup::Dict{Int, Int}
     branch_lookup::Dict{String, Int}
+    transformer_3w_lookup::Dict{String, Vector{String}}
```
GabrielKS, line 87: What is this? Explain in the docstring

src/common.jl
```diff
@@ -75,7 +75,11 @@ function _get_withdrawals!(
     bus_lookup::Dict{Int, Int},
     sys::PSY.System,
 )
-    loads = PSY.get_components(x -> !isa(x, PSY.FixedAdmittance), PSY.ElectricLoad, sys)
+    loads = PSY.get_components(
+        x -> (!isa(x, PSY.FixedAdmittance) && !isa(x, PSY.SwitchedAdmittance)),
```
GabrielKS, line 79: This would require changes to PSY and might not be your problem but maybe it would be good to put `FixedAdmittance` and `SwitchedAdmittance` under a common supertype if they're going to be filtered on together like this?

GabrielKS (changes requested):
Are we planning to use this `transformer_3w_lookup` data for something?

src/common.jl
jd-lara: @GabrielKS lets open a PR to review the type hierarchy in PSY and update before we merge this code

src/PowerFlowData.jl
rbolgaryn: We need a way to map the branch results back to the 3W transformer component. The 3W transformer is an AC branch but is represented by 3 branches in the calculation at the data level and Ybus level. So if we need to know the results for the 3W transformer, we need to backtrace them from the 3 branches.

jd-lara:
@rbolgaryn what's the status of this branch?

rbolgaryn:
> @rbolgaryn what's the status of this branch?

I will check if it is up to date with PSY 5 once it gets released. Otherwise should be fine.

jd-lara:
@rbolgaryn  I am going to close this because now this implementation is not going to be valid given the changes in PNM.

PR #123, opened 2025-03-21
Bugfix and test for "large residual" warning

### LINKED ISSUE #108: Out-of-range bus index in "large initial residual" warning
In `ac_power_flow.jl`, the `ix` of the ["too large initial residual" warning](https://github.com/NREL-Sienna/PowerFlows.jl/blob/main/src/ac_power_flow.jl#L82) may be larger than the number of buses, leading to a `KeyError`. I don't have a system that reliably produces this, but if you add `@assert issubset(1:length(res), keys(data.bus_lookup))` at line 81, that assertion fails.

luke-kiernan:
Fix #108 and add a test case that checks that the "large residual at bus" warning is indeed about the correct bus.


src/powerflow_method.jl
```diff
@@ -330,11 +330,12 @@ function _newton_powerflow(
     J(time_step)  # we need to fill J with values because at this point it was just initialized
 
     if sum(abs, residual.Rv) > WARN_LARGE_RESIDUAL * length(residual.Rv)
-        n_buses = first(size(data.bus_type))
         _, ix = findmax(residual.Rv)
-        bx = ix <= n_buses ? ix : ix - n_buses
-        bus_no = data.bus_lookup[bx]
-        @warn "Initial guess provided results in a large initial residual. Largest residual at bus $bus_no"
+        pow_type = ix % 2 == 1 ? "active" : "reactive"
+        bus_ix = div(ix + 1, 2)
```
GabrielKS, line 335: nitpick: `divrem` could get you both of these

test/test_iterative_methods.jl
```diff
@@ -36,3 +36,18 @@ end
...
+    sys = PSB.build_system(PSB.PSISystems, "RTS_GMLC_DA_sys")
+    data = PowerFlowData(ACPowerFlow(), sys)
+    # First, write solution to data. Then set magnitude of a random bus to a huge number
+    # and try to solve again: the "large residual warning" should be about that bus.
+    solve_powerflow!(data)
+    bus = rand(collect(PSY.get_components(PSY.ACBus, sys)))
+    bus_no = bus.number
+    bus_ix = PowerFlows.get_bus_lookup(data)[bus_no]
+    data.bus_magnitude[bus_ix] = 100.0
+    @test_logs (:warn, Regex(".*Largest residual at bus $(bus_no).*")
```
GabrielKS, line 51:
```suggestion
    @test_logs (:warn, r".*Largest residual at bus $(bus_no).*"
```

test/test_iterative_methods.jl
```diff
@@ -36,3 +36,18 @@ end
...
 # need jacobian to be ill-conditioned...
+
+@testset "large residual warning" begin
+    # a system where bus numbers aren't 1, 2,...is there a smaller one with this property?
+    sys = PSB.build_system(PSB.PSISystems, "RTS_GMLC_DA_sys")
+    data = PowerFlowData(ACPowerFlow(), sys)
+    # First, write solution to data. Then set magnitude of a random bus to a huge number
+    # and try to solve again: the "large residual warning" should be about that bus.
+    solve_powerflow!(data)
+    bus = rand(collect(PSY.get_components(PSY.ACBus, sys)))
```
GabrielKS, line 47: I like my tests to be deterministic, maybe hardcode an index to perturb? Or if you want more than that, loop over the test a few times with random bus based on a hardcoded random seed

jd-lara:
@luke-kiernan can you revise this PR?

src/powerflow_method.jl
```diff
@@ -330,11 +330,12 @@ function _newton_powerflow(
...
+        bus_no = axes(data.power_network_matrix, 1)[bus_ix]
+        @warn "Initial guess provided results in a large initial residual. " *
+              "Largest residual at bus $bus_no ($bus_ix by matrix indexing; $pow_type power)"
```
GabrielKS, line 338: Would it be useful to put the value of the residual here too? (No problem if not)

GabrielKS (commented):
Looks pretty good, a few small comments

test/test_iterative_methods.jl
luke-kiernan: I tried that first, it doesn't work because $ has special meaning in regex, so I went to the [Julia forum](https://discourse.julialang.org/t/regex-w-string-interpolation/3545) which suggests the above solution.

src/powerflow_method.jl
luke-kiernan: Sure, but I find it clearer as-is: if it was 0-indexed I could just do `pow_type, bus_ix = ix % 2, ix / 2`, but it's 1-indexed...

test/test_iterative_methods.jl
GabrielKS: Oh, nice

GabrielKS (approved):
LGTM!

Issue #107, opened 2025-02-24
Dependency on KLU.jl private interface


GabrielKS:
As of https://github.com/NREL-Sienna/PowerFlows.jl/pull/97, we are relying on significant private interface from [KLU.jl](https://github.com/JuliaSparse/KLU.jl). We should figure out a long-term plan for how to eliminate this unsafe dependency.

GabrielKS:
From synchronous discussion with @jd-lara et al., the current plan is to use a `[compat]` entry to pin the KLU.jl dependency to the exact release we tested PowerFlows with. Possible options to consider in the longer term include (1) writing our own KLU wrapper and (2) benchmarking to see if using only public interface of KLU.jl is really that much worse.

luke-kiernan:
This has come up again, so it seems time to address it. Options:

1. Submit a PR to `KLU.jl` that eliminates the big allocations by their public interface. If I recall correctly, most of them come from one issue: they're using a function (`_klu_name`) to pick the correct C function at runtime based on the types of the index/value, when they ought to use multiple dispatch to do so at compile time.
2. Instead of having `rf_{numeric/symbolic/common}` as fields, write `Ref(K.[thing])` where we need those things. (Still mildly dependent on the private interface.) Or remove the depency on the private interface completely, and allocate/deallocate via `setfield!`.

luke-kiernan:
I didn't end up opening up an issue, but I did do a performance comparison. I wrote a simplified version of `KLULinSolveCache` that only uses the public interface. HEre's the benchmark for the simplified one:

![Image](https://github.com/user-attachments/assets/b5b964bc-4c29-42a0-940d-439ad390f1d0)
And here's the benchmark for the current, more optimized one:

![Image](https://github.com/user-attachments/assets/8999309c-fc52-484a-adcd-b0a3cdfddf32)
Comparing the medians, the current one is about 5% faster. That's not a whole lot. I'll repeat this experiment on a larger system (e.g. CATS) on Monday.

PR #125, opened 2025-03-24
DC powerflow fallback for large residual

luke-kiernan:
Added a `robust_power_flow` field to `ACPowerFlow`: if created like
```
pf = ACPowerFlow{NewtonRaphsonACPowerFlow}(; robust_power_flow = true)
```
and the residual at initial guess is large, then `_newton_powerflow` will run a DC powerflow to try to find a better initial guess.


test/test_powerflow_data.jl
```diff
@@ -30,27 +30,34 @@ end
     data_modified = PowerFlowData(ACPowerFlow{ACSolver}(), sys_original)
     modify_rts_powerflow!(data_modified)
 
+    # BUG/FIXME: exclude = [:aux_network_matrix] here is a patchwork solution.
+    # IS.compare_values tries to access the ABA matrix at a one-too-large index, likely
+    # due to the fact that the slack bus is omitted. Is the bug in IS, PNM, or PF, though?
```
GabrielKS, line 35: Let's not adopt the patchwork solution until we're sure that it is actually a bug upstream that is causing this (and even then, it would be best to fix that now so we don't have to use the patchwork solution). Have you verified manually that the old `aux_network_matrix` is the same as the new one?

GabrielKS (changes requested):
Implementation looks pretty good. Some comments on testing.

test/test_powerflow_data.jl
GabrielKS: I might be convinced to let a compromise stopgap through where we exclude `aux_network_matrix` from `IS.compare_values` but then compare it ourselves manually. But really it would be great to fix this bug now.

luke-kiernan: This should go away soon: I've tracked down the bug and issued a PR [here](https://github.com/NREL-Sienna/PowerNetworkMatrices.jl/pull/105). Rodrigo has given it his ok, so it'll probably be merged tomorrow

src/powerflow_method.jl
```diff
@@ -282,12 +282,16 @@ function _newton_powerflow(
     time_step::Int64;
     kwargs...) where {T <: Union{TrustRegionACPowerFlow, NewtonRaphsonACPowerFlow}}
     residual = ACPowerFlowResidual(data, time_step)
-    x0 = calculate_x0(data, time_step)
+    if pf.robust_power_flow
+        x0::Vector{Float64} = improved_x0(data, time_step, residual)
```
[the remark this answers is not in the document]
luke-kiernan: My goal here is to make sure the compiler knows that `x0` will be of the same type, regardless of whether it's defined in the `if` or in the `else`.

src/definitions.jl
```diff
@@ -2,8 +2,8 @@ const MAX_INIT_RESIDUAL = 10.0
 const BOUNDS_TOLERANCE = 1e-6
 const MAX_REACTIVE_POWER_ITERATIONS = 10
 const DEFAULT_MAX_REDISTRIBUTION_ITERATIONS = 10
-const WARN_LARGE_RESIDUAL = 10 # print a warning about "bad initial guess" if 
-# sum(abs, residual)/length(residual) > 10.
+const LARGE_RESIDUAL = 10 # threshold for "bad initial guess": default
+# norm(residual, 1)/length(residual) > 10.
```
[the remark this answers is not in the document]
luke-kiernan: The difference is purely semantic: `norm(x,1)` equals `sum(abs, x)`. (Look up L1 norm.) I just thought that `norm` in this case is clearer, because we're looking at size of something.

src/powerflow_method.jl
```diff
@@ -352,3 +384,21 @@ function calculate_x0(data::ACPowerFlowData,
...
+# flows. For AC fallback, we're only interested in the current timestep, and no branch flows
+function _dc_powerflow_fallback!(data::ACPowerFlowData, time_step::Int)
+    # PERF: if multi-period and multiple time steps have bad initial guesses,
+    #       we're re-creating this factorization for each time step. Store it inside
+    #       data.aux_network_matrix instead.
+    ABA_matrix = data.aux_network_matrix.data
+    solver_cache = KLULinSolveCache(ABA_matrix)
+    full_factor!(solver_cache, ABA_matrix)
+    p_inj = data.bus_activepower_injection[data.valid_ix, time_step]
+    -data.bus_activepower_withdrawals[data.valid_ix, time_step]
```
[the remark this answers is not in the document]
luke-kiernan: For some reason, the formatter insists on doing things this way. I'll experiment, see if I leave it as one line if it picks a different indentation itself.

src/powerflow_method.jl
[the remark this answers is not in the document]
GabrielKS: Could investigate this with `@code_warntype`, but I'm pretty sure it already knows, since, within `improved_x0`, `x0` always comes from `calculate_x0`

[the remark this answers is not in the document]
luke-kiernan: Yeah looks like it's fine without.

test/test_iterative_methods.jl
```diff
@@ -18,6 +18,46 @@ end
...
+    # TODO: case with bad residual where DC powerflow doesn't yield improvement?
+
+    # check that it does the DC fallback.
+    for (i, comp) in enumerate(get_components(PSY.PowerLoad, sys2))
+        set_max_active_power!(comp, 100.0)
+        set_active_power!(comp, 100.0)
+    end
+    improvement_regex = r".*DC powerflow fallback yields better x0"
+    @test_logs (:info, improvement_regex) match_mode = :any PF.solve_powerflow(pf, sys2)
+end
```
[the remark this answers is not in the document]
luke-kiernan: I can't come up with a simple, robust way to test the difference in behavior. I could try to cook up some initial conditions such that the powerflow converges in fewer iterations when `robust_power_flow = true`, but that's a fairly brittle test. If our default powerflow improves, then suddenly that test will break. Would you include that anyway?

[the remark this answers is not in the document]
GabrielKS: Yes — don't do the brittle thing and do do the log message thing.

GabrielKS (approved):
Looks good aside from excluding `aux_network_matrix` in testing. Can we get https://github.com/NREL-Sienna/PowerNetworkMatrices.jl/pull/105 merged and commit to releasing it before this so we don't have to merge this with that exclusion in place?

src/PowerFlowData.jl
```diff
@@ -231,15 +231,19 @@ function PowerFlowData(
...
     converged = fill(false, time_steps)
+    # PERF: type instability.
     # loss factors order matches the order of buses in the grid model, and is calculated for all buses including ref buses (equals 0 for ref buses)
     if calculate_loss_factors
         loss_factors = Matrix{Float64}(undef, (n_buses, length(timestep_names)))
     else
         loss_factors = nothing
     end
-
+    if pf.robust_power_flow
```
jd-lara, line 242: use a getter function here

src/powerflow_method.jl
```diff
@@ -283,11 +283,16 @@ function _newton_powerflow(
     kwargs...) where {T <: Union{TrustRegionACPowerFlow, NewtonRaphsonACPowerFlow}}
     residual = ACPowerFlowResidual(data, time_step)
     x0 = calculate_x0(data, time_step)
+    if pf.robust_power_flow
```
jd-lara, line 286: This check should come after the check for the residual size

src/powerflow_method.jl
```diff
@@ -352,3 +386,22 @@ function calculate_x0(data::ACPowerFlowData,
...
     return x0
 end
+
+"""When solving AC power flows, if the initial guess has large residual, we run a DC power 
+flow as a fallback. This runs a DC powerflow on `data::ACPowerFlowData` for the given
+`time_step`, and writes the solution to `data.bus_angles`."""
+function _dc_powerflow_fallback!(data::ACPowerFlowData, time_step::Int)
+    # dev note: for DC, we can efficiently solve for all timesteps at once, and we want branch
+    # flows. For AC fallback, we're only interested in the current timestep, and no branch flows
+    # PERF: if multi-period and multiple time steps have bad initial guesses,
```
jd-lara, line 396: Resolve this before merge

jd-lara (approved):
This PR looks of to me as a first pass for the robust power flow approach

src/powerflow_method.jl
luke-kiernan: Ok, I re-arranged it to
```
if [large residual] && [do DC fallback]
    run dc fallback
    recalculate residual
end

if [(still) large residual]
   print warning
end
```

src/powerflow_method.jl
luke-kiernan: I'm assuming you mean the "PERF" comment, not the "dev note" (which is an aside on why I structured things this way). There's 2 larger things tied up in this.

First, there's the switch from the public interface of `KLU.jl` to our own `KLULinSolveCache`, for performance and flexibility reasons. However, there's still some concerns with that switch: see issue #107.

Second, the `KLUFactorization` optionally stored by `ABA_Matrix` isn't used anywhere in PF currently. As far as PF is concerned, it could simply be removed...but I might that factorization after all, right here. Is that good enough reason to have `ABA_Matrix` store a factorization, though?

Once those 2 design decisions (`KLUFactorization` vs `KLULinSolveCache`, whether it's worth having `ABA_Matrix` store a factorization just for the DC fallback) are made, it'll be clearer has to how to address this.

luke-kiernan:
Was hoping to resolve issue #107 before merging this, so that some of the DC powerflow stuff could be tidied up. But it's been 2 weeks, and there's other PRs in the pipeline.

PR #134, opened 2025-03-27
implement distributed slack for 1 REF bus and 1 connected component

rbolgaryn:
- PowerFlowData receives a 2D array for slack participation factors at every bus and time step

- by default set to 0 except for REF buses (logic must be revisited when implementing for multiple REF buses)

- slack participation factors can be passed as a kwarg to AC solve_powerflow functions:
    *  nothing - default behavior of 1 for REF bus
    * Dict{String, Float64} - names of generators as keys and values of the slack participations factors as values
    * Vector{Dict{String, Float64}} - for multiperiod calculation, each entry in the vector corresponds to a separate time step.

- adding a test for several configurations of slack participation factors


test/test_solve_powerflow.jl
```diff
@@ -549,3 +549,109 @@ end
...
 end
+
+@testset "AC PF with distributed slack" begin
+    sys = PSB.build_system(PSB.PSITestSystems, "c_sys14"; add_forecasts = false)
+    bus_numbers = get_bus_numbers(sys)
+
+    original_gen_power = zeros(Float64, 14)
+    with_units_base(sys, UnitSystem.NATURAL_UNITS) do
+        original_gen_power .= [
+            isempty(g) ? 0 : get_active_power(first(g)) for g in [
```
GabrielKS, line 560:
```suggestion
            isempty(g) ? 0 : get_active_power(only(g)) for g in [
```

GabrielKS (commented):
Is this going to be something people will want to use in power flow in the loop? If so, we'll need to design that interface.

src/ac_power_flow_residual.jl
```diff
@@ -100,47 +145,66 @@ Calling the `ACPowerFlowResidual` will also update the values of P, Q, V, Θ in
...
     ::Val{PSY.ACBusTypes.PV})
     # When bustype == PV PSY.Bus, state variables are Reactive Power Generated and Voltage Angle
+    P_net[ix] = P_net_set[ix] + P_slack
     Q_net[ix] = StateVector[2 * ix - 1]
+    data.bus_activepower_injection[ix, time_step] =
+        P_net[ix] + data.bus_activepower_withdrawals[ix, time_step]
     data.bus_reactivepower_injection[ix, time_step] =
         StateVector[2 * ix - 1] + data.bus_reactivepower_withdrawals[ix, time_step]
     data.bus_angles[ix, time_step] = StateVector[2 * ix]
 end
```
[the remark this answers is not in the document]
rbolgaryn: I can just call the other function and then add the one line for bus angle

src/solve_ac_powerflow.jl
```diff
@@ -198,6 +198,9 @@ function _ac_powerflow(
     kwargs...,
 )
     check_reactive_power_limits = get(kwargs, :check_reactive_power_limits, false)
+    if :slack_participation_factors in keys(kwargs)
```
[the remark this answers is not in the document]
rbolgaryn: It allows passing the kwargs as slack participation factors to the power flow function, and only having one change in the code. This interface is not final and can change e.g. by having a dict instead. So I chose not to invest much effort in it while the final interface is not yet clear.

src/ac_power_flow_residual.jl
```diff
@@ -43,21 +46,53 @@ function ACPowerFlowResidual(data::ACPowerFlowData, time_step::Int64)
...
+    bus_type = view(data.bus_type, :, time_step)
     for ix in 1:n_buses
         P_net[ix] =
             data.bus_activepower_injection[ix, time_step] -
             data.bus_activepower_withdrawals[ix, time_step]
         Q_net[ix] =
             data.bus_reactivepower_injection[ix, time_step] -
             data.bus_reactivepower_withdrawals[ix, time_step]
+        P_net_set[ix] = P_net[ix]
+        if bus_type[ix] ∈ (PSY.ACBusTypes.REF, PSY.ACBusTypes.PV)
```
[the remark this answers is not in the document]
rbolgaryn: changed to use dispatch on bus type

src/ac_power_flow_residual.jl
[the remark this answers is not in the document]
rbolgaryn: changed to use a common function to set p and q

src/solve_ac_powerflow.jl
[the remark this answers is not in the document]
rbolgaryn: changed it to Dict{ACBus, Float64} as a kwarg for ACPowerFlow -> this way it can be also considered in PowerSimulations

test/test_solve_powerflow.jl
```diff
@@ -549,3 +549,123 @@ end
...
+                get_components(x -> get_number(get_bus(x)) == i, ThermalStandard, sys)
+                for i in bus_numbers
+            ]
+        ]
+    end
+
+    pf = ACPowerFlow()
+    res1 = solve_powerflow(pf, sys)
+
+    ref_b = []
```
luke-kiernan, line 570: Nitpick: do you really need 4 different lists here? I think you could probably just use bit vectors (one per bus type) instead: e.g. `ref_b = buses[ref_mask]` and `ref_n = bus_numbers[ref_mask]`

luke-kiernan (changes requested):
Few comments, might have more tomorrow. Also, it'd be good to have a test where the results aren't just different, but known (e.g. something highly symmetric).

luke-kiernan (approved):
Looks good to me (but I'll wait for Gabriel's ok to merge)

src/common.jl
```diff
@@ -209,6 +209,109 @@ function make_dc_powerflowdata(
...
 
+function make_bus_slack_participation_factors(
+    sys::System,
+    generator_slack_participation_factors::Dict{String, Float64},
+    bus_lookup::Dict{Int, Int},
+    time_steps::Int,
+    n_buses::Int,
+    ::Matrix{PSY.ACBusTypes},
+)
+    I = Int[]
```
GabrielKS, line 220: Style nitpick: non-constant variable names should not start with a capital letter

src/PowerFlowData.jl
```diff
@@ -89,6 +89,12 @@ struct PowerFlowData{
     bus_activepower_withdrawals::Matrix{Float64}
     bus_reactivepower_withdrawals::Matrix{Float64}
     bus_reactivepower_bounds::Matrix{Vector{Float64}}
+    generator_slack_participation_factors::Union{
+        Dict{String, Float64},
+        Vector{Dict{String, Float64}},
```
GabrielKS, line 94: I think our convention so far has been to use the same type whether it's single time period or multi time period and just have one slice along the time axis if it's single time period (see all the `Matrix` fields here). Is there a reason that doesn't work here?

src/ac_power_flow_residual.jl
```diff
@@ -40,24 +46,68 @@ Create an instance of `ACPowerFlowResidual` for a given time step.
 and net bus reactive power injections.
 """
 function ACPowerFlowResidual(data::ACPowerFlowData, time_step::Int64)
+    is_ref(bt::Val{PSY.ACBusTypes.REF}) = true
```
GabrielKS, line 49: Unconventional to put functions inside of functions

src/common.jl
GabrielKS: Never mind, seems justifiable in this case

src/common.jl
```diff
@@ -209,6 +209,113 @@ function make_dc_powerflowdata(
...
+        val == 0.0 && continue
+        gen = PSY.get_component(PSY.Generator, sys, gen_name)
+        isnothing(gen) && throw(ArgumentError("Generator $gen_name not found"))
+        PSY.get_available(gen) || continue
+        bus = PSY.get_bus(gen)
+        bus_idx = bus_lookup[PSY.get_number(bus)]
+        for time_step in 1:time_steps
+            push!(I, bus_idx)
+            push!(J, time_step)
+            push!(V, val)
```
GabrielKS, line 234: Does this correctly handle the case where multiple generators are attached to one bus?

src/common.jl
```diff
@@ -209,6 +209,113 @@ function make_dc_powerflowdata(
...
+
+    for (time_step, factors) in enumerate(generator_slack_participation_factors)
+        for (gen_name, val) in factors
+            val == 0.0 && continue
+            gen = PSY.get_component(PSY.Generator, sys, gen_name)
+            isnothing(gen) && throw(ArgumentError("Generator $gen_name not found"))
+            PSY.get_available(gen) || continue
+            bus = PSY.get_bus(gen)
+            push!(I, bus_lookup[PSY.get_number(bus)])
+            push!(J, time_step)
```
GabrielKS, line 281: Code duplication…

src/post_processing.jl
```diff
@@ -465,7 +465,12 @@ function write_powerflow_solution!(
         elseif bus.bustype == PSY.ACBusTypes.PV
             Q_gen = data.bus_reactivepower_injection[ix, time_step]
             bus.angle = data.bus_angles[ix, time_step]
-            _reactive_power_redistribution_pv(sys, Q_gen, bus, max_iterations)
+            if data.bus_slack_participation_factors[ix, time_step] != 0.0
```
GabrielKS, line 468: Might be nice to have a comment explaining this change

src/common.jl
```diff
@@ -209,6 +209,113 @@ function make_dc_powerflowdata(
...
+    n_buses::Int,
+    ::Matrix{PSY.ACBusTypes},
+)
+    I = Int[]
+    J = Int[]
+    V = Float64[]
+
+    for (gen_name, val) in generator_slack_participation_factors
+        val == 0.0 && continue
+        gen = PSY.get_component(PSY.Generator, sys, gen_name)
```
GabrielKS, line 226: You are technically allowed to have two generators with the same name as long as they don't have the same type, so this isn't guaranteed to return only one component. Maybe the keys of `generator_slack_participation_factors` should be tuples of data type, string?

src/powerflow_types.jl
```diff
@@ -8,21 +8,46 @@ struct ACPowerFlow{ACSolver <: ACPowerFlowSolverType} <: PowerFlowEvaluationMode
     check_reactive_power_limits::Bool
     exporter::Union{Nothing, PowerFlowEvaluationModel}
     calculate_loss_factors::Bool
+    generator_slack_participation_factors::Union{
+        Nothing,
+        Dict{String, Float64},
```
GabrielKS, line 13: See above about how this is not a unique way of referring to generators

src/PowerFlowData.jl
GabrielKS: Or maybe if the interpretation of `Dict{String, Float64}` is "same factors for all time periods," you should just create a vector with the same data in all entries in the constructor? That would also save you from having so many `make_bus_slack_participation_factors` implementations

src/powerflow_types.jl
```diff
@@ -8,21 +8,46 @@ struct ACPowerFlow{ACSolver <: ACPowerFlowSolverType} <: PowerFlowEvaluationMode
...
+        Vector{Dict{String, Float64}},
```
GabrielKS, line 14: You might decide to change the typing in `PowerFlowData` to just `Union{Nothing, Vector{Dict{String, Float64}}` but keep the `Dict{String, Float64}` option here. That seems like a fine user interface.

test/test_solve_powerflow.jl
```diff
@@ -549,3 +549,178 @@ end
...
+
+@testset "AC PF with distributed slack" for (grid_lib, grid_name) in [
+    (PSB.PSITestSystems, "c_sys14"),
+    (PSB.MatpowerTestSystems, "matpower_case30_sys"),
+]
+    function _get_spf_dict(slack_participation_factors)
+        generator_slack_participation_factors = Dict{String, Float64}()
+        for (b, spf) in enumerate(slack_participation_factors)
+            get_bustype(
+                only(get_components(x -> get_number(x) == bus_numbers[b], ACBus, sys)),
```
GabrielKS, line 561:
```suggestion
                get_bus(sys, bus_numbers[b])
```

test/test_solve_powerflow.jl
```diff
@@ -549,3 +549,178 @@ end
...
+        bus_numbers,
+        original_bus_power,
+        original_gen_power,
+        data_original_bus_power,
+    )
+
+    # TODO: add a test when two generators at the same bus have unequal slack participation factors
+    # TODO: add test for multiperiod with just one time step and one Dict input in the vector
+    # TODO: add test for multiperiod with just one Dict input of slack factors for different time steps
+    # TODO: add test for multiperiod with different Dict inputs for different time steps
```
GabrielKS, line 725: Should these be addressed now?

test/test_utils/common.jl
```diff
@@ -81,3 +81,94 @@ function modify_rts_powerflow!(data::PowerFlowData)
...
 end
+
+function _check_distributed_slack_consistency(
+    gen_power::Vector{Float64},
+    slack_participation_factors::Vector{Float64},
+    original_bus_power::Vector{Float64},
+)::Bool
+    slack_provided = gen_power .- original_bus_power
+    nnz = slack_participation_factors .!= 0.0
+    test_1 = all(isapprox.(slack_provided[.!nnz], 0.0, atol = 1e-6, rtol = 0))
```
GabrielKS, line 92: If it would be useful, you can use `@test` in functions like this — that might get you more granular test failures

GabrielKS (changes requested):

src/PowerFlowData.jl
GabrielKS: One more option: if the only purpose of `generator_slack_participation_factors` is to allow the creation of `bus_slack_participation_factors`, do you need to store them in the struct at all?

src/ac_power_flow_residual.jl
GabrielKS: not sure what this does to performance

src/PowerFlowData.jl
rbolgaryn: Thank you!

I added a possibility to just have 1 input for all time steps if the factors are same, so that not to store unnecessarily the same data multiple times.

The intended sequence is:

Generator factors -> Bus factors -> Bus active power results -> Generator active power results.

I implemented the generator factors to be stored in data for this purpose.

src/common.jl
rbolgaryn: Yes, because of the way a sparse matrix is initialized - duplicates are summed. 

TODO: add comment in the code for clarity

src/ac_power_flow_residual.jl
rbolgaryn: TODO: remove the functions is_ref, is_refpv  after the @scoped_enum is improved

This PR already addresses it: https://github.com/NREL-Sienna/InfrastructureSystems.jl/pull/458

src/PowerFlowData.jl
rbolgaryn: OK, if the dictionaries are the same in the vector, then they are stored by reference

rbolgaryn: Modified as suggested in a different comment: support all three variants in the ACPowerFlow kwargs, but support 2 in PowerFlowData - vector and Nothing.

luke-kiernan (approved):

test/test_solve_powerflow.jl
rbolgaryn: yes, these tests have been implemented

src/common.jl
```diff
@@ -209,37 +209,61 @@ function make_dc_powerflowdata(
...
+    J::Vector{Int},
+    V::Vector{Float64},
     sys::System,
-    generator_slack_participation_factors::Dict{String, Float64},
+    gsp_factors::Dict{Tuple{DataType, String}, Float64},
     bus_lookup::Dict{Int, Int},
-    time_steps::Int,
-    n_buses::Int,
-    ::Matrix{PSY.ACBusTypes},
+    time_steps_iter::UnitRange{Int},
```
GabrielKS, line 219: can this be `AbstractVector{<:Integer}`?

src/post_processing.jl
```diff
@@ -164,9 +164,12 @@ function _power_redistribution_ref(
     Q_gen::Float64,
     bus::PSY.Bus,
     max_iterations::Int,
+    gspf::Union{Nothing, Dict{Tuple{DataType, String}, Float64}},
 )
     devices_ =
         PSY.get_components(x -> _is_available_source(x, bus), PSY.StaticInjection, sys)
+    all_devices = devices_
+
     sources = filter(x -> typeof(x) == PSY.Source, collect(devices_))
```
GabrielKS, line 173:
```suggestion
    sources = filter(x -> x isa PSY.Source, collect(devices_))
```

src/post_processing.jl
```diff
@@ -164,9 +164,12 @@ function _power_redistribution_ref(
     Q_gen::Float64,
     bus::PSY.Bus,
     max_iterations::Int,
+    gspf::Union{Nothing, Dict{Tuple{DataType, String}, Float64}},
```
GabrielKS, line 167: What is this? Need not do a full docstring for every helper function but would be nice to define non-obvious abbreviations

src/post_processing.jl
```diff
@@ -457,17 +484,27 @@ function write_powerflow_solution!(
...
             P_gen = data.bus_activepower_injection[ix, time_step]
             Q_gen = data.bus_reactivepower_injection[ix, time_step]
-            _power_redistribution_ref(sys, P_gen, Q_gen, bus, max_iterations)
+            _power_redistribution_ref(sys, P_gen, Q_gen, bus, max_iterations, gspf)
         elseif bus.bustype == PSY.ACBusTypes.PV
             Q_gen = data.bus_reactivepower_injection[ix, time_step]
             bus.angle = data.bus_angles[ix, time_step]
+            # If the PV bus has a nonzero slack participation factor, 
+            # then not only reactive power but also active power could have been changed 
+            # in the power flow calculation. This requires thge same
```
GabrielKS, line 503:
```suggestion
            # in the power flow calculation. This requires the same 
```

src/post_processing.jl
```diff
@@ -457,17 +484,27 @@ function write_powerflow_solution!(
...
             Q_gen = data.bus_reactivepower_injection[ix, time_step]
-            _power_redistribution_ref(sys, P_gen, Q_gen, bus, max_iterations)
+            _power_redistribution_ref(sys, P_gen, Q_gen, bus, max_iterations, gspf)
         elseif bus.bustype == PSY.ACBusTypes.PV
             Q_gen = data.bus_reactivepower_injection[ix, time_step]
             bus.angle = data.bus_angles[ix, time_step]
+            # If the PV bus has a nonzero slack participation factor, 
+            # then not only reactive power but also active power could have been changed 
+            # in the power flow calculation. This requires thge same 
+            # active and reacvtive power redistribution step as for the REF bus.
```
GabrielKS, line 504:
```suggestion
            # active and reactive power redistribution step as for the REF bus.
```

test/test_multiperiod_ac_powerflow.jl
```diff
@@ -119,3 +119,60 @@ end
     solve_powerflow!(data_brute_force; pf = pf)
     @test isnothing(data_brute_force.loss_factors)
 end
+
+@testset failfast = true "MULTI-PERIOD power flows evaluation with DS" for mode in (
```
GabrielKS, line 123: why do you want to `failfast`?

test/test_multiperiod_ac_powerflow.jl
```diff
@@ -119,3 +119,60 @@ end
...
+    :equal,
+    :dict,
+    :array_1,
+    :array_24,
+)
+    # get system
+    sys = PSB.build_system(PSB.PSITestSystems, "c_sys14"; add_forecasts = false)
+    generators = get_components(ThermalStandard, sys)
+    create_gspf(generators) =
+        Dict((ThermalStandard, get_name(x)) => abs(randn()) for x in generators)
```
GabrielKS, line 134: use a random seed for reproducibility, see e.g. https://github.com/NREL-Sienna/InfrastructureSystems.jl/pull/458

test/test_multiperiod_ac_powerflow.jl
rbolgaryn: I have @test in a for loop for 24 time steps:

```Julia
for time_step in 1:time_steps
        _check_distributed_slack_consistency(
            data.bus_activepower_injection[:, time_step],
            collect(data.bus_slack_participation_factors[:, time_step]),
            init_p_injections[:, time_step],
        )
    end
```

If it fails once, there is no additional benefit for waiting for it to fail more times. and having all the tracebacks in the terminal.

src/post_processing.jl
rbolgaryn: renamed to a readable variable name

test/test_multiperiod_ac_powerflow.jl
rbolgaryn: https://github.com/NREL-Sienna/InfrastructureSystems.jl/pull/457

rbolgaryn: I added the random seed, but we need to wait for #458 to be merged to IS first

rbolgaryn: I am using 0 for the test as seed

GabrielKS (approved):

jd-lara:
DNM until I take a look please

src/ac_power_flow_residual.jl
```diff
@@ -43,21 +49,60 @@ function ACPowerFlowResidual(data::ACPowerFlowData, time_step::Int64)
...
-    for ix in 1:n_buses
+
+    P_net_set = zeros(Float64, n_buses)
+    bus_type = view(data.bus_type, :, time_step)
+
+    spf_idx = Int[]
+    spf_val = Float64[]
+    sum_sl_weights = 0.0  # for scope
+
+    # slack_bus is set to the first REF bus found - will be used for the total slack power
```
jd-lara, line 60: We need to check that there is one Ref Bus per subnetwork only really.

src/ac_power_flow_residual.jl
```diff
@@ -43,21 +49,60 @@ function ACPowerFlowResidual(data::ACPowerFlowData, time_step::Int64)
...
+    # TODO: check if there are multiple REF buses and treat the remaining ones as if they were PV buses
+    # TODO: enable multiple slack buses, multiple disconnected zones
```
jd-lara, line 62: We need to open an issue for islanded power flow solutions

jd-lara (changes requested):
The PR looks good overall. There are some inconsistencies to check for the existence of Ref vs slacks that needs to be considered correctly please review that.

PR #139, opened 2025-04-07
Trust Region: Singular J's

luke-kiernan:
I implemented a fallback for singular J's for the trust region and made a single helper function `_set_Δx_nr!` that calculates the Newton-Raphson step for both methods. Also, now the `Δx_nr` field is used to store the Newton-Raphson step in both methods.

### DIFF
--- a/src/powerflow_method.jl
+++ b/src/powerflow_method.jl
@@ -1,9 +1,8 @@
 """Cache for non-linear methods
 # Fields
--`x::Vector{Float64}`: the current state vector. Used for all methods.
--`r::Vector{Float64}`: the current residual. Used for all methods.
-    For `NewtonRaphsonACPowerFlow`, we solve `J_x Δx = r` in-place,
-    so this also stores the step `Δx` at times in those methods.
+-`x::Vector{Float64}`: the current state vector.
+-`r::Vector{Float64}`: the current residual.
+-`Δx_nr::Vector{Float64}`: the step under the Newton-Raphson method.
 The remainder of the fields are only used in the `TrustRegionACPowerFlow`:
 -`r_predict::Vector{Float64}`: the predicted residual at `x+Δx_proposed`,
     under a linear approximation: i.e `J_x⋅(x+Δx_proposed)`.
@@ -14,8 +13,7 @@ The remainder of the fields are only used in the `TrustRegionACPowerFlow`:
     is inside and `x+Δx_nr` outside. The dogleg step selects the point where the line
     from `x+Δx_cauchy` to `x+Δx_nr` crosses the boundary of the trust region.
 -`Δx_cauchy::Vector{Float64}`: the step to the Cauchy point if the Cauchy point
-    lies within the trust region, otherwise a step in that direction.
--`Δx_nr::Vector{Float64}`: the step under the Newton-Raphson method."""
+    lies within the trust region, otherwise a step in that direction."""
 struct StateVectorCache
     x::Vector{Float64}
     r::Vector{Float64} # residual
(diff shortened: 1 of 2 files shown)


src/powerflow_method.jl
```diff
@@ -35,20 +33,91 @@ function StateVectorCache(x0::Vector{Float64}, f0::Vector{Float64})
     return StateVectorCache(x, r, r_predict, Δx_proposed, Δx_cauchy, Δx_nr)
 end
 
+"""Returns a stand-in matrix for singular J's."""
+function _singular_J_fallback(Jv::SparseMatrixCSC{Float64, Int32},
+    x::Vector{Float64})
+    fjac2 = Jv' * Jv
+    lambda = NR_SINGULAR_SCALING * sqrt(length(x) * eps()) * norm(fjac2, 1)
```
GabrielKS, line 40: Out of curiosity, what is the middle factor here for?

src/powerflow_method.jl
```diff
@@ -35,20 +33,91 @@ function StateVectorCache(x0::Vector{Float64}, f0::Vector{Float64})
...
+end
+
+"""Check error and do refinement."""
+function _do_refinement!(stateVector::StateVectorCache,
+    A::SparseMatrixCSC{Float64, Int32},
+    cache::KLULinSolveCache{Int32},
+    refinement_threshold::Float64,
+    refinement_eps::Float64,
+)
+    # use stateVector.r_predict as temporary buffer.
```
GabrielKS, line 59: nice

src/powerflow_method.jl
```diff
@@ -35,20 +33,91 @@ function StateVectorCache(x0::Vector{Float64}, f0::Vector{Float64})
...
+    if delta > refinement_threshold
+        stateVector.Δx_nr .= solve_w_refinement(cache,
+            A,
+            stateVector.r,
+            refinement_eps)
+    end
+    return
+end
+
+"""Sets the Newton-Raphson step. Usually, this is just `J.Jv \\ stateVector.r`, but
```
GabrielKS, line 73: `\\`?

GabrielKS (commented):
Code looks good, haven't reviewed the math. Is it possible to test this behavior, even just finding a case that would error/be obviously wrong before and isn't now?

src/powerflow_method.jl
luke-kiernan: Do I not need to escape the backslash?

luke-kiernan:
About the behvavior: I was going thru the large systems in the old "PowerNetworkMatrices_performance_and_tests" repo, running `@benchmark` on one of the variants of `solve_powerflow`. One system failed to converge with NR: when I switched to the TrustRegion method, it returned an error for singular J. Which is what brought this to my attention.

src/powerflow_method.jl
GabrielKS: Oh right, nah this is fine.

test/test_iterative_methods.jl
```diff
@@ -34,3 +34,14 @@ end
...
     end
 end
+
+@testset "singular Jacobian trust region" begin
+    # NewtonRaphsonACPowerFlow fails to converge on this system.
+    pf = ACPowerFlow{TrustRegionACPowerFlow}()
+    sys = build_system(MatpowerTestSystems, "matpower_ACTIVSg10k_sys")
+    data = PowerFlowData(pf, sys)
+    @test_logs (:warn, Regex(".*Jacobian is singular.*")
+    ) match_mode = :any for _ in 1:20
```
GabrielKS, line 44: The rationale here is that by solving the same data a bunch of times serially, we create some sort of degeneracy that ends up invoking the singular Jacobian handler, right? Maybe add a comment to that effect

GabrielKS (approved):
LGTM pending comment and passing CI

test/test_iterative_methods.jl
luke-kiernan: Sure, I can add such a comment. It's completely random, though--there's no reason to expect this to happen. It's just the example I tripped up.

PR #145, opened 2025-04-10
Rb/ds2

rbolgaryn:
Implements distributed slack for multiple connected components

### DIFF
--- a/src/ac_power_flow_residual.jl
+++ b/src/ac_power_flow_residual.jl
@@ -61,7 +61,8 @@ function ACPowerFlowResidual(data::ACPowerFlowData, time_step::Int64)
     # TODO: check if there are multiple REF buses and treat the remaining ones as if they were PV buses
     # TODO: enable multiple slack buses, multiple disconnected zones
 
-    subnetworks = _find_subnetworks_for_reference_buses(data.power_network_matrix.data, bus_type)
+    subnetworks =
+        _find_subnetworks_for_reference_buses(data.power_network_matrix.data, bus_type)
 
     for (ix, bt) in zip(1:n_buses, bus_type)
         P_net[ix] =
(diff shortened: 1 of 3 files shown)


src/ac_power_flow_residual.jl
```diff
@@ -276,27 +290,33 @@ function _update_residual_values!(
...
-            Q_net,
-            P_net_set,
-            P_slack[ix],
-            x,
-            data,
-            time_step,
-            Val(bt),
-        )
+
+    for (slack_bus, subnetwork_buses) in subnetworks
```
GabrielKS, line 301: This isn't looping through all `ACBusTypes.REF` buses `slack_bus`, just the `first` one in each subnetwork, right? Is that arbitrariness a problem?

src/ac_power_flow_residual.jl
```diff
@@ -328,3 +348,25 @@ function _update_residual_values!(
     end
     return
 end
+
+function _find_subnetworks_for_reference_buses(
```
GabrielKS, line 352: Here we are basically taking the output of `PNM.find_subnetworks`, which is a dictionary from the first bus in a subnetwork to a set of all the buses in the subnetwork for all subnetworks, and building a similar structure except the keys are reference buses, right? So not every reference bus gets an entry if there are multiple in a given subnetwork?

test/test_solve_powerflow.jl
```diff
@@ -907,3 +618,145 @@ end
     )
     reset_p()
 end
+
+@testset "AC PF DS with two connected components" for (mode, gen_mode) in (
```
GabrielKS, line 622:
```suggestion
@testset "AC PF DS with two connected components" for mode in (:same, :random), gen_mode in (:gen, :source)
```

test/test_solve_powerflow.jl
```diff
@@ -907,3 +618,145 @@ end
...
+        pf2,
+        sys,
+        bus_slack_participation_factors,
+        bus_numbers,
+        original_bus_power,
+        original_gen_power,
+        data_original_bus_power;
+        check_connectivity = false,
+    )
+end
```
GabrielKS, line 762: Test the case of multiple `ACBusTypes.REF` buses in a given subnetwork, also test that it errors properly when there are zero `ACBusTypes.REF` buses in a given subnetwork

test/test_utils/common.jl
```diff
@@ -158,14 +177,171 @@ function _check_ds_pf(
...
+"""
+    _check_name(sys::System, name::String, component_type::DataType)
+    Check if the name is unique in the system. If not, append a number to the name.
+"""
+function _check_name(sys::System, name::String, component_type::DataType)
+    # Check if the name is unique
+    check = true
+    i = 1
+    while check
+        existing_components_with_name = get_component(component_type, sys, name)
```
GabrielKS, line 203: You can use `has_component`

GabrielKS (changes requested):
Some questions about the case of multiple reference buses per subnetwork. I like the streamlining of the tests!

luke-kiernan (approved):
This is just bringing the branch up to speed with `main`, correct? The way you've integrated the singular J stuff looks straightfoward and reasonable. (So ok to merge by me, pending Gabriel's remarks and tests passing.)

PR #148, opened 2025-04-14
Lk/robust homotopy

luke-kiernan:


luke-kiernan:
The first time the CI ran, it failed for Windows, with a "failure to pre-compile `MUMPS.jl`" message. I tried adding `MPIPreferences.jl` as a dependency (but didn't write `using MPIPreferences`) and pushed my changes: then the CI failed on Mac, with 
```
Run julia-actions/setup-julia@latest
  with:
    version: 1
    arch: x64
    include-all-prereleases: false
    show-versioninfo: false
Warning: [setup-julia] x64 arch has been requested on a macOS runner that has an arm64 (Apple Silicon) architecture. You may have meant to use the "aarch64" arch instead (or left it unspecified for the correct default).
Download of https://julialang-s3.julialang.org/bin/mac/x64/1.11/julia-1.11.5-mac64.tar.gz failed, trying again. Error: Error: Unexpected HTTP response: 403
Download of https://julialang-s3.julialang.org/bin/mac/x64/1.11/julia-1.11.5-mac64.tar.gz failed, trying again. Error: Error: Unexpected HTTP response: 403
[...]
Error: Unexpected HTTP response: 403
```
My best guess of what's going on here: the CI isn't playing nicely with the way `MPI.jl` detects which backend to use.

luke-kiernan:
Update on the Windows bug: I discovered that the sequential, MPI-free version of `MUMPS.jl` is also broken on Windows, but not on Mac. This leads me to think the problem lies in the `build_tarballs.jl` scripts in Yggdrasil. i.e. the Julia code itself is fine, but the binary artifacts `MUMPS_jll` and `MUMPS_seq_jll` aren't. `BinaryBuilder.jl` emits a couple warnings during the build process, which could point in the right direction. See the above linked issue for details. However, I'm going to let this bug rest for now.

luke-kiernan:
Here's some options:
1. Use `Pardiso.jl` on Windows and `MUMPS.jl` on Mac/Linux. That'd require maintaining two separate backends, though.
2. Use [LDLFactorizations.jl](https://github.com/JuliaSmoothOptimizers/LDLFactorizations.jl) However, the second sentence in its readme basically says "don't expect this to be competitive with `MUMPS.jl`, `HSL.jl`, or `Pardiso.jl`."
3. Use `HSL.jl`. Fast and cross-platform, but requires a license and nontrivial build/setup process on Windows (or so says ChatGPT).

luke-kiernan:
I've rebased onto psy5 and opened a separate PR. Leaving this branch intact for now, just in case.

Issue #138, opened 2025-04-03
Handle large initial residuals using Levenberg-Marquardt


jd-lara:
Try the strategy of using Levenberg-Marquardt before the call for NR. Even though this method can be slower, it can make it easier to converge for NR.

luke-kiernan:
Removed link because reading the above, I realize I'm not really doing what you asked: I'm not using L-M before the call to NR, but instead leaving it as a separate method.

Well...actually...should I fix that now, in PR I just unlinked above, or in a separate one? I guess I have a bit of PTSD from the first PR that snowballed into an unmanageable size.

PR #149, opened 2025-04-16
Levenberg-Marquardt implementation

luke-kiernan:
A working implementation of Levenberg-Marquardt. See issue #138. There's a few details to chase down here, but it's 95% of the way there.

One point of ickiness: I need to be able to do `A .+= B' * B` without changing the sparse structure of `A`. If I merely run that code snippet, `SparseArrays` automatically drops the numerically zero entries. My current workaround (in `A_plus_eq_BT_B!`) relies on the fact that `SparseArrays` doesn't drop the numerically zero entries when computing `M = B' * B`. That seems iffy. Perhaps there's some lower-level routine I can call (e.g. LAPACK/BLAS), where the spec explicitly says "this does not drop numerically zero entries."

edit: I opened [an issue](https://github.com/JuliaSparse/SparseArrays.jl/issues/614) in SparseArrays.jl


test/test_levenburg_marquardt.jl
```diff
@@ -0,0 +1,126 @@
...
+    end
+    dense_graph = (5, denseEdges)
+    multi_component = (6, [(1, 2), (2, 3), (4, 5)])
+    # start with a loop of size 9, add in a few other edges
+    complex_graph_edges = Vector{Tuple{Int, Int}}([(i, i + 1) for i in 1:8])
+    push!(complex_graph_edges, (9, 1))
+    append!(complex_graph_edges, [(3, 9), (1, 4), (3, 5), (6, 8)])
+    complex_graph = (9, complex_graph_edges)
+
+    graphs = [chain_graph, dense_graph, multi_component, complex_graph]
```
luke-kiernan, line 115: oh, oops. I forgot the loop graph here.

test/test_levenburg_marquardt.jl
```diff
@@ -0,0 +1,126 @@
...
+    return Set(zip(row, col))
+end
+
+function testSameStructure(A1::SparseMatrixCSC, A2::SparseMatrixCSC)
+    @test A1.rowval == A2.rowval && A1.colptr == A2.colptr
+end
+
+@testset "A_plus_eq_BT_B!" begin
+    # test basic numeric behavior.
+    B = sprand(100, 100, 0.01)
```
luke-kiernan, line 12: Set the RNG seed.

test/test_levenburg_marquardt.jl
```diff
@@ -0,0 +1,127 @@
...
+    firstCol = 1
+    while all(A[:, firstCol] .== 0.0)
+        firstCol += 1
+    end
+    vals = SparseArrays.nonzeros(B)
+    for i in SparseArrays.nzrange(B, firstCol)
+        vals[i] = 0.0
+    end
+    temp = B' * B
+    @assert (firstCol, firstCol) in coordSet(temp)
```
GabrielKS, line 39: Just confirming that this is properly `@assert` and not `@test`?

GabrielKS (commented):
Some minor comments, haven't examined the math in detail.

src/levenberg-marquardt.jl
```diff
@@ -0,0 +1,212 @@
...
+    # J' * J is dot products of pairs of columns.
+    # so look at pairs of columns and check if there's a row in which both are nonzero.
+    # i.e. look at pairs of buses and see if they have a neighbor in common.
+    rows = Int32[]      # I
+    columns = Int32[]   # J
+    values = Float64[]  # V
+
+    # an over-estimate: this counts directed paths of 2 edges. Might be many 
+    # paths of length 2 between the same pair of points that share a neighbor.
+    numEdgePairs = sum(x -> length(x)^2, get_branch_lookup(data); init = 0)
```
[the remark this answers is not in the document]
luke-kiernan: Good catch. It should be `get_neighbor`, not `get_branch_lookup`: I want the sum of the squares of the degrees of the vertices.

test/test_levenburg_marquardt.jl
luke-kiernan: The logic of this test case rests on some mildly iffy assumptions about how SparseArrays.jl works under the hood. Here, I'm checking those assumptions: I'm testing the behavior of my testing code, not the behavior of `PowerFlows.jl`, which is why I did `@assert`. Still a little weird, though

src/levenberg-marquardt.jl
```diff
@@ -0,0 +1,212 @@
...
+    sizehint!(rows, 4 * numEdgePairs)
```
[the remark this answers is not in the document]
luke-kiernan: The difference is 2.5% for 2k bus system.

jd-lara:
@luke-kiernan in powernetwork matrices I have a sparse matrix addition that's optimized

src/levenberg-marquardt.jl
```diff
@@ -0,0 +1,212 @@
+"""Driver for the LevenbergMaquardtACPowerFlow method: sets up the data 
+structures (e.g. residual), runs the powerflow method via calling `_run_powerflow_method` 
+on them, then handles post-processing (e.g. loss factors)."""
+function _newton_powerflow(
```
jd-lara, line 4: Why are we calling this newton if it's using LM? This isn't the robust approach

luke-kiernan: It's for multiple dispatch reasons, to match the call [here](https://github.com/NREL-Sienna/PowerFlows.jl/blob/main/src/solve_ac_powerflow.jl#L203). We could re-name it to something else, though: is there a word for "Newton-esque?" Newtonian?

luke-kiernan: "This isn't the robust approach" not sure what you mean there. I'm aware that I've implemented a full, separate LM method, when in the issue, you say "handle large residual via L-M."  However, debugging and code structure wise, it seemed simpler to make a full L-M method first, and integrate it as a fallback strategy second.

luke-kiernan:
The convergence is pretty fragile...but I'm not sure if it's my code or just the limitations of the method itself. (And if it's the limitations of the method itself, which enhancements will help?) We need a better experimental framework. I'd like to be able to measure: 
1.  **Robustness of convergence**. Pick a sensible random starting point with difficulty `d`, run the power flow, and see if it converges. Repeat and look at aggregate stats as we adjust `d`. Most basic form: take a step of size `ϵ` from the solution in a random direction.
2. **Progress of the iterative method**. There's a lot of things that can cause issues: local minima, narrow canyons, too small/large initial step size, too aggressive/conservative in adjusting the step size, etc. Most basic form: at each time step, log the search parameters plus some basic stats (step size, angle with previous step, expected and actual residual improvement).
3. **Performance**, perferrably in combination with the above. Maybe use [`BenchmarkGroup`](https://juliaci.github.io/BenchmarkTools.jl/stable/manual/#The-BenchmarkGroup-type)?

luke-kiernan:
Update on where things stand: solving `(J' J + λ I) Δx = -J' r` is equivalent to finding the least-squares solution to `vcat(J, λ^(1/2) I) Δx = vcat(-r, 0)`. The second is usually chosen because it's more numerically stable, though the first is sometimes faster. After switching to the least-squares one, convergence is much more reliable.

Unfortunately, that matrix isn't square, so we can't use our current KLU-based structs for caching the symbolic factorization. The upside is that sparsity stuff works better: there's no need for a structure-preserving `A .+= B' B`, and `vcat(J, λ^(1/2) I)` has fewer nonzero entries than `J' J + λ I`.

luke-kiernan:
The amount of merges here got out of hand, and then a rebase lead to duplicate commits. To avoid this messiness, I just rebased the final commit onto psy5. I'll leave this branch up for now, in case the history turns out to be relevant, but I'll close this PR.

PR #154, opened 2025-04-23
Jacobian Performance

luke-kiernan:
Some performance enhancements with the Jacobian. So far, I've cut the runtime of `_update_jacobian_matrix_values!` in half (on a 10k bus system), by changing how we compute the terms where `bus_from` and `bus_to` are the same. It's at the cost of some very small allocations--4 `Base.RefValue{Float}`'s in each iteration.

 I'm going to see if I can do more, so leaving this in draft for now.


luke-kiernan:
The test that's failing: my tweak made it so J is no longer singular in a certain test case. When I wrote that test case, I knew it was inherently unstable and could easily break in the future. I'll probably just remove it for now and create an issue as a reminder to find a reliable way to cover the "J is singular" code in a test case.

src/ac_power_flow_jacobian.jl
```diff
@@ -90,7 +90,7 @@ J(time_step)  # Updates the Jacobian matrix stored internally in J (J.Jv) with t
 J.Jv  # Access the Jacobian matrix stored internally in J.
 ```
 """
-function ACPowerFlowJacobian(data::ACPowerFlowData, time_step::Int64)
+function ACPowerFlowJacobian(data::ACPowerFlowData, time_step::Int64, old = false)
```
jd-lara, line 93: @luke-kiernan what's the purpose of this?

luke-kiernan: oops, leftover from debugging

jd-lara (approved):

jd-lara:
@luke-kiernan fix this conflict to merge please

PR #155, opened 2025-04-24
Rb/simplified_hvdc

rbolgaryn:


src/common.jl
jd-lara, line 1: Why does this matter. The DCBuses are for MultiTerminal HVDC

rbolgaryn: We get an error otherwise

jd-lara: Why? and what's the error these checks for types usually give me the ick

GabrielKS (commented):
If the objection is just one of code style, there's the multiple dispatch alternative:
```julia
_get_bustype_wrapper(::PSY.Bus) = missing  # since `nothing` is a possible option for `ACBus.bustype`
_get_bustype_wrapper(bus::PSY.ACBus) = PSY.get_bustype(bus)

bt = PSY.get_bustype(bus)
ismissing(bt) && continue
```

But I agree in general that we're not super great at thoughtfully handling the `DCBus` case across PowerFlows (maybe across Sienna?). We should probably solve that problem more holistically.

jd-lara:
> If the objection is just one of code style, there's the multiple dispatch alternative:
> 
> ```julia
> _get_bustype_wrapper(::PSY.Bus) = missing  # since `nothing` is a possible option for `ACBus.bustype`
> _get_bustype_wrapper(bus::PSY.ACBus) = PSY.get_bustype(bus)
> 
> bt = PSY.get_bustype(bus)
> ismissing(bt) && continue
> ```
> 
> But I agree in general that we're not super great at thoughtfully handling the `DCBus` case across PowerFlows (maybe across Sienna?). We should probably solve that problem more holistically.

Why would we ever get a DCBus in this context? It is an impossibility for this approach, there has to be something more fundamentally wrong.

rbolgaryn:
This was the case giving me the issues:

```Julia
sys = build_system(PSISystems, "sys10_pjm_ac_dc")
julia> data = PowerFlowData(ACPowerFlow(), sys, check_connectivity=false)


ERROR: MethodError: no method matching get_bustype(::PowerSystems.DCBus)
The function `get_bustype` exists, but no method is defined for this combination of argument types.

Closest candidates are:
  get_bustype(::PowerSystems.ACBus)
   @ PowerSystems ~/repos/PowerSystems.jl/src/models/generated/ACBus.jl:108

Stacktrace:
 [1] _initialize_bus_data!(bus_type::Vector{…}, bus_angles::Vector{…}, bus_magnitude::Vector{…}, temp_bus_map::Dict{…}, bus_lookup::Dict{…}, sys::PowerSystems.System)
   @ PowerFlows ~/repos/PowerFlows.jl/src/common.jl:122
 [2] make_powerflowdata(sys::PowerSystems.System, time_steps::Int64, power_network_matrix::PowerNetworkMatrices.Ybus{…}, aux_network_matrix::Nothing, n_buses::Int64, n_branches::Int64, bus_lookup::Dict{…}, branch_lookup::Dict{…}, temp_bus_map::Dict{…}, branch_type::Vector{…}, timestep_map::Dict{…}, valid_ix::Vector{…}, neighbors::Vector{…}, converged::Vector{…}, loss_factors::Nothing, calculate_loss_factors::Bool)
   @ PowerFlows ~/repos/PowerFlows.jl/src/common.jl:234
 [3] PowerFlowData(pf::ACPowerFlow{NewtonRaphsonACPowerFlow}, sys::PowerSystems.System; time_steps::Int64, timestep_names::Vector{String}, check_connectivity::Bool)
   @ PowerFlows ~/repos/PowerFlows.jl/src/PowerFlowData.jl:243
 [4] top-level scope
   @ REPL[10]:1
Some type information was truncated. Use `show(err)` to see complete types.
```

rbolgaryn:
So you can see how the DC bus is there and causing the initialization of the PowerFlowData to crash

jd-lara:
> So you can see how the DC bus is there and causing the initialization of the PowerFlowData to crash. In the test case in PSI I ended up using instead, with TwoTerminalGenericHVDCLine, it does not occur. Nevertheless, it crashing when DC buses are present is still a bug.

The correct fix is to change the places where we cal `get_bus` to ACBus instead of Bus... this is come legacy code before we added DCBuses. 

image

rbolgaryn:
> > So you can see how the DC bus is there and causing the initialization of the PowerFlowData to crash. In the test case in PSI I ended up using instead, with TwoTerminalGenericHVDCLine, it does not occur. Nevertheless, it crashing when DC buses are present is still a bug.
> 
> The correct fix is to change the places where we cal `get_bus` to ACBus instead of Bus... this is come legacy code before we added DCBuses.
> 
> image

I removed all the mentions of DCBus

GabrielKS (approved):
Works for me if we're sure this all will only ever need to work with `ACBus`

jd-lara:
> Works for me if we're sure this all will only ever need to work with `ACBus`

Yes, I mean there is no AC PowerFlow within DC Buses

jd-lara (approved):

PR #158, opened 2025-04-29
fix docstring to reflect new solve_powerflow

yasirroni:
Minor fox doc string


luke-kiernan (approved):
Thanks for catching this!

PR #160, opened 2025-04-29
Separate state vector, data updates from "compute residual" code

luke-kiernan:
Make some helper functions, so that we don't need to index the state vector directly as often. Most of this is just moving code around.

`update_data!`: in the current code, if I want to update the fields of `data` based on the values of `x`, I'd call `residual(x, time_step)`. That's pretty clunky and undesireable.  (I may not have an `ACPowerFlowResidual` object on hand, or I might not want to change `residual.Rv`.)

`partition_state`: mostly intended for debugging, making infographics, etc.

`update_state!`: allows for inspecting `x` during or after the iterative method, without changing the source code and re-compiling. Let's say I want to compute some stats about progress of the iterative method, regarding `x` as a mathematical vector. Now there's an interface that allows for that.


GabrielKS (approved):
LGTM

PR #169, opened 2025-05-16
Add tests for new psy 5 components

luke-kiernan:
For each component that's new or changed significantly, run a powerflow on a system that contains said component.

### DIFF
--- a/src/common.jl
+++ b/src/common.jl
@@ -75,7 +75,11 @@ function _get_withdrawals!(
     bus_lookup::Dict{Int, Int},
     sys::PSY.System,
 )
-    loads = PSY.get_components(x -> !isa(x, PSY.FixedAdmittance), PSY.ElectricLoad, sys)
+    loads = PSY.get_components(
+        x -> !isa(x, PSY.FixedAdmittance) && !isa(x, PSY.SwitchedAdmittance),
+        PSY.ElectricLoad,
+        sys,
+    )
     for l in loads
         !PSY.get_available(l) && continue
         bus = PSY.get_bus(l)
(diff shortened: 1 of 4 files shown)


src/common.jl
```diff
@@ -75,7 +75,11 @@ function _get_withdrawals!(
     bus_lookup::Dict{Int, Int},
     sys::PSY.System,
 )
-    loads = PSY.get_components(x -> !isa(x, PSY.FixedAdmittance), PSY.ElectricLoad, sys)
+    loads = PSY.get_components(
```
jd-lara, line 78: In this case it depends on the control mode of the switched admittance. We need to get this filters in a separate function probably and have more complex queries.

test/test_new_components.jl
```diff
@@ -0,0 +1,26 @@
...
+@testset "new psy 5 components" begin
+    # TODO are there other components that should be included here?
+    componentToSys = Dict{Type, Tuple{Type, String}}(
+        TapTransformer => (MatpowerTestSystems, "matpower_case2_sys"),
+        TModelHVDCLine => (PSISystems, "sys10_pjm_ac_dc"),
+        TwoTerminalVSCLine => (PSSEParsingTestSystems, "pti_vsc_hvdc_test_sys"),
+        SwitchedAdmittance => (PSSEParsingTestSystems, "pti_case24_sys"),
+        TwoTerminalLCCLine =>
+            (PSSEParsingTestSystems, "pti_two_terminal_hvdc_test_sys"),
+        DiscreteControlledACBranch =>
```
jd-lara, line 11: This one will have to be absorbed into the Ybus calculations

luke-kiernan: Yeah, I should've looked more closely at the errors: it's in PowerNetworkMatrices, not here in PowerFlows.

PR #170, opened 2025-05-22
adding voltage stability factors calculation

rbolgaryn:
Adding voltage stability index following the method from the publication below:

P. -A. Lof, T. Smed, G. Andersson, and D. J. Hill, "Fast calculation of a voltage stability index," in IEEE Transactions on Power Systems, vol. 7, no. 1, pp. 54-64, Feb. 1992, doi: 10.1109/59.141687.

The $\sigma$ value corresponding to $G_s$ is written in the position of the REF bus. Singular vector $v$ values indicate the sensitivity of the buses, and are written in the positions of the PQ buses. Values of $v$ for PV buses are set to 0.


src/ac_power_flow_jacobian.jl
```diff
@@ -497,3 +497,148 @@ function calculate_loss_factors(
...
+
+# Returns
+- `rows::Vector{Int32}`: Row indices for the block Jacobian matrix.
+- `cols::Vector{Int32}`: Column indices for the block Jacobian matrix.
+"""
+function block_J_indices(pvpq::Vector{<:Integer}, pq::Vector{<:Integer})
+    rows = Int32[]
+    cols = Int32[]
+
+    for i in pvpq
```
luke-kiernan, line 555: You don't need the loops:
```
rows = vcat(2 .* pvpq .- 1, 2 .* pq)
cols = vcat(2 .* pvpq, 2 .* pq .- 1)
```

src/ac_power_flow_jacobian.jl
```diff
@@ -497,3 +497,148 @@ function calculate_loss_factors(
...
+- `v::Vector{Float64}`: The estimated right singular vector.
+"""
+function find_sigma_uv(
+    Jv::SparseMatrixCSC{Float64, Int32},
+    npvpq::Integer;
+    tol::Float64 = 1e-9,
+    max_iter::Integer = 100,
+)
+    f = KLU.klu(Jv)
+    ft = KLU.klu(sparse(transpose(Jv)))
```
luke-kiernan, line 597: No need to separately calculate a factorization for the transpose: `KLU.jl` has built-in functionality such that you can transpose the factorization as a whole and do things like `transpose(f) \ v`.

src/ac_power_flow_jacobian.jl
```diff
@@ -497,3 +497,148 @@ function calculate_loss_factors(
...
+function calculate_voltage_stability_factors(
+    data::ACPowerFlowData,
+    J::ACPowerFlowJacobian,
+    time_step::Integer,
+)
+    ref, pv, pq = PowerFlows.bus_type_idx(data, time_step)
+    pvpq = [pv; pq]
+    npvpq = length(pvpq)
+    rows, cols = PowerFlows.block_J_indices(pvpq, pq)
+    σ, u, v = PowerFlows.find_sigma_uv(J.Jv[rows, cols], npvpq)
```
luke-kiernan, line 528: Allocating a second big matrix here is probably avoidable. There are some permutations matrices `P1` and `P2` such that dropping the last 2 rows and columns of `P1 J.Jv P2` gives you `Jp = J.Jv[rows, cols]`. Could you just pass `P1`, `P2`, and `J.Jv` to `find_sigma_uv`?

luke-kiernan: Ideally, you'd also only use a KLU factorization of `J.Jv`, but I'm not sure that's feasible. If it was just a permutation, it'd be fine, but dropping those 2 rows and columns messes with things.

rbolgaryn: Seems to require quite a bit of effort to avoid allocations here, I suggest not doing it right now.

jd-lara: I prefer you do a good implementation now. If you know how to do it well don't do it badly.

rbolgaryn: I don't think we know now how to do it without allocation because we would still need to drop the columns/rows for the method to work, and it would be allocating still.

src/ac_power_flow_jacobian.jl
```diff
@@ -497,3 +497,139 @@ function calculate_loss_factors(
...
+- `data::ACPowerFlowData`: The instance containing the grid model data.
+- `J::ACPowerFlowJacobian`: The Jacobian matrix cache.
+- `time_step::Integer`: The calculated time step.
+"""
+function calculate_voltage_stability_factors(
+    data::ACPowerFlowData,
+    J::ACPowerFlowJacobian,
+    time_step::Integer,
+)
+    ref, pv, pq = PowerFlows.bus_type_idx(data, time_step)
```
GabrielKS, line 524: nitpick: no need to do `PowerFlows.` within PowerFlows

src/powerflow_method.jl
```diff
@@ -364,6 +364,9 @@ function _newton_powerflow(
         if data.calculate_loss_factors
             calculate_loss_factors(data, J.Jv, time_step)
         end
+        if data.calculate_voltage_stability_factors
+            calculate_voltage_stability_factors(data, J, time_step)
```
GabrielKS, line 368: Best to not have two different things both called `calculate_voltage_stability_factors`

src/ac_power_flow_jacobian.jl
```diff
@@ -497,3 +497,139 @@ function calculate_loss_factors(
...
+The function uses the method described in the following publication:
+
+    P.-A. Lof, T. Smed, G. Andersson, and D. J. Hill, "Fast calculation of a voltage stability index," in IEEE Transactions on Power Systems, vol. 7, no. 1, pp. 54-64, Feb. 1992, doi: 10.1109/59.141687.
+
+# Arguments
+- `data::ACPowerFlowData`: The instance containing the grid model data.
+- `J::ACPowerFlowJacobian`: The Jacobian matrix cache.
+- `time_step::Integer`: The calculated time step.
+"""
+function do_calculate_voltage_stability_factors(
```
jd-lara, line 519: Why is this function `do_`? we don't use that pattern anywhere.

jd-lara (commented):
@rbolgaryn  please rebase this branch onto PSY5 and lets try to get it merged asap

src/ac_power_flow_jacobian.jl
```diff
@@ -521,3 +521,139 @@ function calculate_loss_factors(
...
+# Keyword Arguments
+- `tol::Float64=1e-6`: Convergence tolerance for the iterative algorithm.
+- `max_iter::Integer=100`: Maximum number of iterations.
+
+# Returns
+- `σ::Float64`: The estimated smallest singular value.
+- `u::Vector{Float64}`: The estimated left singular vector.
+- `v::Vector{Float64}`: The estimated right singular vector.
+"""
+function find_sigma_uv(
```
jd-lara, line 607: This function needs names for the variables... I can't tell what's happening here to make a meaningful review.

src/ac_power_flow_jacobian.jl
```diff
@@ -521,3 +521,139 @@ function calculate_loss_factors(
...
+    v[d_ix] .= 0.0
+    v ./= norm(v, 2)
+
+    u = ones(n)
+    u[d_ix] .= 0.0
+
+    σ = 1e6
+    k = 1
+
+    while k <= max_iter
```
jd-lara, line 627: This function is leaving a lot of optimizations for matrix algebra out of the table. It needs optimization or open issues about this

src/ac_power_flow_jacobian.jl
```diff
@@ -521,3 +521,139 @@ function calculate_loss_factors(
...
+        u .= transpose(f) \ v
+        u[d_ix] .= 0.0
```
jd-lara, line 629: why are you using a `.=` for a single index write.

src/ac_power_flow_jacobian.jl
```diff
@@ -521,3 +521,139 @@ function calculate_loss_factors(
...
+        d_σ = σ_1 - σ
+        σ = σ_1
+
+        u ./= n_u
+
+        if abs(d_σ) < tol
+            break
+        end
+
+        v .= f \ u
```
jd-lara, line 642: same here why `v`

src/ac_power_flow_jacobian.jl
```diff
@@ -521,3 +521,139 @@ function calculate_loss_factors(
...
+        v[d_ix] .= 0.0
+        n_v = norm(v, 2)
+
+        σ_2 = 1 / n_v
+        d_σ = σ_2 - σ
+        σ = σ_2
+
+        v ./= n_v
```
jd-lara, line 650: and again here

jd-lara (changes requested):
This PR is almost there but the actual method's function needs refactoring to make it readable and improve the linear algebra performance. @luke-kiernan you can help @rbolgaryn with the Linear Algebra tricks

src/ac_power_flow_jacobian.jl
rbolgaryn: add comment "in the paper it is ..."

luke-kiernan:
> This PR is almost there but the actual method's function needs refactoring to make it readable and improve the linear algebra performance. @luke-kiernan you can help @rbolgaryn with the Linear Algebra tricks

I tried a couple different things, but didn't see significant improvement. If this is high priority, I have ideas, but they'd require diving into the details of the linear algebra. (The core idea: if we select the pivoting order in a certain way, we'd get an LU factorization of the submatrix `J[pvpq, pvpq]` for free.)

jd-lara (approved):

PR #171, opened 2025-05-22
Add Vm Validation and autoscaling to Trust Region Method

luke-kiernan:
I realized that the trust region method is converging to nonsense solutions on the large `ACTIVSg10k` test system:
```
using PowerFlows, PowerSystemCaseBuilder
sys = build_system(MatpowerTestSystems, "matpower_ACTIVSg10k_sys")
pf_tr = ACPowerFlow{TrustRegionACPowerFlow}()
data_tr = PowerFlowData(pf_tr, sys)
solve_powerflow!(data_tr; pf = pf_tr, maxIterations = 200, factor = 0.1)
v = deepcopy(data_tr.bus_magnitude[:, 1])
sort!(v)
display(v)
```
All entries of `v` should be between 0.9 and 1.1, yet under the current branch, that is far from the case. In this PR, I add the following:

1. Log a warning when the voltage magnitude for any PQ bus leaves the 0.5-1.5 range. 
2. Re-normalize the different components in the TrustRegion method. I followed `NLSolve.jl` and set `d[i] = max(0.1 * d[i], norm(J[:, i]))`: then we do the math like we're minimzing `norm(F(x ./ d))`, instead of `F(x)`. This makes it so all but 2 of the voltage magnitudes above are sensible.
3. The  problematic voltage magnitudes were among those with the smallest `d` values: reasoning that the iterative procedure was taking too big of steps in certain directions, I added a constant factor to the autoscaling weights via `d .+= 10*minimum(d)`.  This resolves those last 2 voltage magnitudes.

That 3rd change was rather arbitrary: I left a `TODO` reminder in the code to look into a more principled solution.


luke-kiernan:
On second thought, maybe we should wait to merge this until there's a test case that compares the result of our powerflow with what matpower or PSS/e gets. I'm only validating the entries corresponding to voltage magnitude: there could be issues with other entries.

GabrielKS:
I fail to reproduce on `psy5`:
9dda1194-0d8e-40e5-a0c2-243bf67b8532

jd-lara:
> I fail to reproduce on `psy5`: 9dda1194-0d8e-40e5-a0c2-243bf67b8532

can we put these into a test please?

luke-kiernan:
I added that test and turned autoscale off by default. Any other changes I should make?

src/common.jl
```diff
@@ -452,3 +452,30 @@ function make_powerflowdata(
...
+    range::NamedTuple{(:min, :max), Tuple{Float64, Float64}} = VM_VALIDATION_RANGE,
+    i::Int64 = 1,
+)
+    outside_range = Vector{Int64}()
+    MAX_INDS_TO_PRINT = 10
+    for (i, bt) in enumerate(bus_types)
+        if bt == PSY.ACBusTypes.PQ
+            if (x[2 * i - 1] < range.min || x[2 * i - 1] > range.max) &&
+               size(outside_range, 1) <= MAX_INDS_TO_PRINT
+                push!(outside_range, i)
```
jd-lara, line 467: allocating inside a hot loop is dangerous. Running this check should be an opt in possibility

luke-kiernan: Good point! I only wrote it that way for convenience: I'll switch to printing out the mis-matching bus numbers as I iterate, instead of storing them. Sound good?

luke-kiernan: Finished: commented it out, opened an issue, and resolved the merge conflicts.

jd-lara (approved):

PR #173, opened 2025-05-30
Throw an error for wrong bus type

luke-kiernan:
Downstream effect of [PR 1398](https://github.com/NREL-Sienna/PowerSystems.jl/pull/1398) in PSY: since we're no longer correcting "wrong bus type" inputs in the parser in PSY (and instead logging an error message), we need to check the bus types here in PF and throw a hard error if they're wrong.

### DIFF
--- a/src/common.jl
+++ b/src/common.jl
@@ -58,7 +58,12 @@ function _get_injections!(
     bus_lookup::Dict{Int, Int},
     sys::PSY.System,
 )
-    sources = PSY.get_components(d -> !isa(d, PSY.ElectricLoad), PSY.StaticInjection, sys)
+    # PSS/E counts buses with switched shunts as voltage-controlled, PV, so we do the same.
+    sources = PSY.get_components(
+        d -> !isa(d, PSY.ElectricLoad) || isa(d, PSY.SwitchedAdmittance),
+        PSY.StaticInjection,
+        sys,
+    )
     has_generator = falses(length(bus_lookup))
     for source in sources
         !PSY.get_available(source) && continue
--- a/src/common.jl
+++ b/src/common.jl
@@ -59,12 +59,34 @@ function _get_injections!(
     sys::PSY.System,
 )
     sources = PSY.get_components(d -> !isa(d, PSY.ElectricLoad), PSY.StaticInjection, sys)
+    has_generator = falses(length(bus_lookup))
     for source in sources
         !PSY.get_available(source) && continue
         bus = PSY.get_bus(source)
+        # Throw an error if we see an available source at a PQ bus.
+        if PSY.get_bustype(bus) == PSY.ACBusTypes.PQ
+            throw(
+                ArgumentError(
+                    "Source $(PSY.get_name(source)) at PQ bus $(PSY.get_name(bus)). Please change the bus type to PV or REF.",
+                ),
+            )
+        end
         bus_ix = bus_lookup[PSY.get_number(bus)]
+        has_generator[bus_ix] = true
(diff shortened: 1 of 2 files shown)


test/test_powerflow_data.jl
```diff
@@ -63,3 +63,46 @@ end
...
 end
+
+"""Helper function that sets availability of all sources at a given bus."""
+function set_availability_at_bus(
+    sys::PSY.System,
+    bus::PSY.ACBus,
+    availability::Bool,
+)
+    for source in
+        PSY.get_components(d -> !isa(d, PSY.ElectricLoad), PSY.StaticInjection, sys)
```
gks-spa, line 74: nitpick: there's a method of `get_components` that can do filtering for you

src/common.jl
```diff
@@ -59,13 +59,35 @@ function _get_injections!(
...
+        end
         bus_ix = bus_lookup[PSY.get_number(bus)]
+        has_generator[bus_ix] = true
         bus_activepower_injection[bus_ix] += PSY.get_active_power(source)
         bus_reactivepower_injection[bus_ix] += PSY.get_reactive_power(source)
     end
+    # Throw an error if there's a PV/REF bus without an available source.
+    for bus in PSY.get_components(PSY.ACBus, sys)
+        bus_ix = bus_lookup[PSY.get_number(bus)]
+        if PSY.get_bustype(bus) != PSY.ACBusTypes.PQ && !has_generator[bus_ix]
```
gks-spa, line 82: nitpick: should this be `!= PQ` or `in (PV, Ref)`? what's the desired behavior if we were to add a new bustype?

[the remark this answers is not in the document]
luke-kiernan: Fair point. I'll change it to `in (PV, REF)`.

test/test_powerflow_data.jl
```diff
@@ -63,3 +63,46 @@ end
...
+        if get_number(get_bus(source)) == get_number(bus)
```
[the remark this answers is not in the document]
luke-kiernan: oh yeah. I could even do it in one line, though it looks pretty awful:
```
set_available!.(
    PSY.get_components(
        d -> !isa(d, PSY.ElectricLoad) && get_number(get_bus(d)) == get_number(source),
        PSY.StaticInjection,
        sys,
    ),
    (availability, )
)
```

src/common.jl
```diff
@@ -59,13 +59,36 @@ function _get_injections!(
...
         bus_ix = bus_lookup[PSY.get_number(bus)]
+        has_generator[bus_ix] = true
         bus_activepower_injection[bus_ix] += PSY.get_active_power(source)
         bus_reactivepower_injection[bus_ix] += PSY.get_reactive_power(source)
     end
+    # Throw an error if there's a PV/REF bus without an available source.
+    for bus in PSY.get_components(PSY.ACBus, sys)
+        bus_ix = bus_lookup[PSY.get_number(bus)]
+        if PSY.get_bustype(bus) in [PSY.ACBusTypes.PV, PSY.ACBusTypes.REF] &&
+           !has_generator[bus_ix]
```
jd-lara, line 83: A PV bus can also have facts devices or switched shunts. This correction needs to be more careful. Consult with @mcllerena on the behavior of these other components.

luke-kiernan: Messasing Marck, PSSE expects a PV bus to satisfy: (has available generator) AND NOT (has unlocked switched shunt/FACTS device). If that is violated, PSSE outputs a warning, but oddly it still runs the powerflow. Should we do the same, or should we throw an error?

jd-lara: Throw and error.

PR #181, opened 2025-06-25
Use Network Reduction

luke-kiernan:
Take into account the network reduction when doing input/output. So far, I've adjusted the initialization of `PowerFlowData`, but there's more to do: reactive power bounds? Output? Branch flows? Some of this may need to wait until PNM's network reduction code is finalized.


luke-kiernan:
I will close this--looks like this should be split between a few different PRs. Also, the commit history is thoroughly confusing due to the rebase-then-merge.

PR #186, opened 2025-07-03
Combine DC power flow solvers via multiple dispatch

luke-kiernan:
Use a type parameter in place of having 3 functions with the same body.

### DIFF
--- a/src/solve_dc_powerflow.jl
+++ b/src/solve_dc_powerflow.jl
@@ -113,78 +113,26 @@ end
 # SINGLE PERIOD ##############################################################
 
 """
-Evaluates the power flows on each system's branch by means of the PTDF matrix.
-Updates the PowerFlowData structure and returns a dictionary containing a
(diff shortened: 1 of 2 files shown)


src/solve_dc_powerflow.jl
```diff
@@ -113,143 +113,47 @@ end
...
-# Arguments:
-- `data::PTDFPowerFlowData`:
-        PowerFlowData structure containing the system data per each timestep
-        considered, as well as the Virtual PTDF matrix.
-- `sys::PSY.System`:
-        container gathering the system data.
-"""
-function solve_powerflow(
-    data::vPTDFPowerFlowData,
+    data::Union{PTDFPowerFlowData, vPTDFPowerFlowData, ABAPowerFlowData},
```
GabrielKS, line 156: use your new abstract type here

GabrielKS (approved):
with one change LGTM

PR #195, opened 2025-07-15
Levenberg-Maquardt 2

luke-kiernan:
The previous PR for the LM method got out of hand, so opening a new one with cleaner commit history.


jd-lara (approved):
LGTM. This is on a pretty workable state.

PR #198, opened 2025-07-15
disable enhanced flat start by default

luke-kiernan:
Only try the "enhanced flat start" strategy if `robust_power_flow` is `true`, and only replace `x0` if that strategy produces a smaller initial residual. (The `TrustRegion` strategy fails to converge on the 10k bus system, if we indiscriminately use the "enhanced flat start" strategy.)

### DIFF
--- a/src/powerflow_method.jl
+++ b/src/powerflow_method.jl
@@ -393,11 +393,19 @@ function _newton_powerflow(
     x0 = calculate_x0(data, time_step)
     residual(x0, time_step)
     if norm(residual.Rv, 1) > LARGE_RESIDUAL * length(residual.Rv) &&
+       get_enhanced_flat_start(pf)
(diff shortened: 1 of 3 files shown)


rbolgaryn (approved):

rbolgaryn (changes requested):
```Julia
if norm(residual.Rv, 1) > LARGE_RESIDUAL * length(residual.Rv)
    enhanced_flat_start(...)
end

if norm(residual.Rv, 1) > LARGE_RESIDUAL * length(residual.Rv) &&
        get_robust_power_flow(pf)
         improve_x0!(x0, data, time_step, residual)
     else
         ...

```

luke-kiernan:
I added a flag so the enhanced flat start can be disabled, but it is on by default. (Otherwise, testing the "large initial residual"  and "dc fallback" code would be difficult.)

PR #200, opened 2025-07-16
fix LM code; move AC pre-solve setup into its own file

luke-kiernan:
Realized I never added the `LevenbergMarquardtACSolver` to the list of solvers that are run during the tests, so a lot of the merged code wasn't actually working. This fixes that.

I also refactored the pre-solve AC setup to use helper functions, and put those functions into a separate file. This way, we have consistent behavior across solvers.

### DIFF
--- a/src/levenberg-marquardt.jl
+++ b/src/levenberg-marquardt.jl
@@ -27,7 +27,7 @@ function _newton_powerflow(
 
     if converged
         @info("The LevenbergMarquardtACPowerFlow solver converged after $i iterations.")
-        loss_factors_helper!(data, x0, residual, J, time_step)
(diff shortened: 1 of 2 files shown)


src/levenberg-marquardt.jl
```diff
@@ -1,40 +1,41 @@
...
+            residual,
+            J;
+            kwargs...,
+        )
+    end
 
     if converged
-        @info("The LevenbergMaquardtACPowerFlow solver converged after $i iterations.")
+        @info("The LevenbergMarquardtACPowerFlow solver converged after $i iterations.")
         loss_factors_helper!(data, x0, residual, J, time_step)
```
jd-lara, line 30: Rename without `_helper` call it `loss_factors_calculation`

PR #201, opened 2025-07-17
Update DC pf tests for branch indexing change in PNM

luke-kiernan:
Branches are now in a different order due to being indexed by tuples of ints, not strings: adjust the DC power flow tests to compensate.

### DIFF
--- a/test/test_dc_powerflow.jl
+++ b/test/test_dc_powerflow.jl
@@ -1,15 +1,19 @@
+function flows_from_dataframe(powerflow_results,
+    branch_lookup::Dict{Tuple{Int, Int}, Int},
+    direction::Symbol = :P_from_to,
+)
+    flow_results_df = powerflow_results["1"]["flow_results"]
+    flows = fill(NaN, length(branch_lookup))
+    for row in eachrow(flow_results_df)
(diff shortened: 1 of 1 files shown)


test/test_dc_powerflow.jl
```diff
@@ -1,16 +1,20 @@
+function flows_from_dataframe(powerflow_results,
+    branch_lookup::Dict{Tuple{Int, Int}, Int},
+    direction::Symbol = :P_from_to,
+)
+    flow_results_df = powerflow_results["1"]["flow_results"]
```
jd-lara, line 5: What's the "1" here?

luke-kiernan: First time step. There is only 1 time step, but the output is still a dictionary of time step label to results.

jd-lara (commented):
@m-bossart this should be a fix in PNM to be able to index by branch and map to the corresponding

m-bossart:
> @m-bossart this should be a fix in PNM to be able to index by branch and map to the corresponding

Right now in PNM we can index by branch name to get the entry for the corresponding arc. Can you give more detail of what else is needed?

PR #207, opened 2025-07-21
Handle Q limits for exporter when nothing value is found

mcllerena:
I address here a problem with the Zone data naming, where its value exceeds the 12 character limit that the manual states. So it is truncated to use the first 12 characters to avoid triggering the assert error.


mcllerena:
> is this PR also included in #214 and we should close it or is it still needed?

Yes, we can closed it. I addressed this issue in the general PR of exporter items.

PR #211, opened 2025-07-24
adding transformer phase shift

rbolgaryn:
Adding transformer phase shift to Transformer2W and TapTransformer as `phase_shift`


luke-kiernan (commented):
Small nitpicks, but logic looks sound.

jd-lara (approved):

PR #212, opened 2025-07-24
Robust Homotopy Performance Improvements

luke-kiernan:
A bunch of RH performance improvements, taking the runtime from 10+ seconds on the 10k bus system to 6 seconds. Inspired by `LinearSolverCache`, I added a layer of abstraction such that we can easily swap out the linear algebra backend: I've implemented a class for KLU, MUMPS, and CHOLMOD's Cholesky. The last is by far the fastest, so I've put that as the default and removed the MUMPS dependencies. This should fix the compilation issues on Windows.


jd-lara (changes requested):
I'll merge to fix the windows compatibility, but we need a more principled approach for the usage of MUMPS. Please open a follow up PR removing all MUMPS code

PR #214, opened 2025-07-25
Handle missing exporter components

mcllerena:
Still under dev, for the PSY side reimport file testing.
- [x]  Facts
- [x]  CircuitBreaker/Switch
- [x]  3W transformers
- [x]  LCC HVDC
- [x]  GenericHVDC (we need to make a warning/message that this device is exported as a source and generator at each end)
- [x]  VSC HVDC
- [x]  Switched Shunts
- [x]  Impedance Correction Tables

Regarding the benchmark when exporting and reading back:
Without Supplemental Attributes
image

With Supplemental Attributes:
image

Does not vary much, although this is just considering the Impedance Correction Data. Will be updated when handling the geoinfo of buses for other files.


src/psse_export.jl
```diff
@@ -90,7 +90,7 @@ mutable struct PSSEExporter <: SystemPowerFlowContainer
                     "PSS/E version $psse_version is not supported, must be one of $PSSE_EXPORT_SUPPORTED_VERSIONS",
                 ),
             )
-        system = PSY.fast_deepcopy_system(base_system)
+        system = PSY.fast_deepcopy_system(base_system; skip_supplemental_attributes = false)
```
GabrielKS, line 93: Did you ever end up testing the performance implications of this? It'd be great if we could get some numbers and add them to the description of this PR for future reference

src/psse_export.jl
```diff
@@ -265,9 +265,21 @@ convert_empty_stringvec = Base.Fix1(convert_empty, Vector{String})
 # PERF could be improved by appending to the buffer rather than doing string interpolation, seems unnecessary
 _psse_quote_string(s::String) = "'$s'"
 
+# Rounds a value up to 4 decimals, to avoid large approximations in file
+_psse_round_val(val) = isa(val, String) ? val : round(val; digits = 4)
```
GabrielKS, line 269: do this with multiple dispatch, also maybe define a const for digits we want to round to

GabrielKS: Also, what's your rule for when this gets used and when it doesn't? Let's be consistent

src/psse_export.jl
```diff
@@ -599,12 +624,13 @@ function write_to_buffers!(
         I = md["bus_number_mapping"][sienna_bus_number]
         ID = _psse_quote_string(load_name_mapping[(sienna_bus_number, PSY.get_name(load))])
         STATUS = PSY.get_available(load) ? 1 : 0
-        AREA = PSSE_DEFAULT  # defaults to bus's area
-        ZONE = PSSE_DEFAULT  # defaults to zone's area
+        AREA = parse(Int, PSY.get_name(PSY.get_area(PSY.get_bus(load))))
```
GabrielKS, line 627: Consider using `_permissive_parse_int`

src/psse_export.jl
```diff
@@ -813,39 +845,61 @@ function write_to_buffers!(
...
         LEN = PSSE_DEFAULT
+        R = PSY.get_r(branch)
+        X = PSY.get_x(branch)
 
-        @fastprintdelim_unroll(io, false, I, J, CKT, R, X, B,
-            RATEA, RATEB, RATEC, GI, BI,
-            GJ, BJ, ST, MET, LEN)
-        fastprintln_psse_default_ownership(io)
+        if branch isa PSY.DiscreteControlledACBranch
+            CKT = replace(BASE_CKT, r"^'|'$" => "")
```
GabrielKS, line 858: Explain what is going on here, why we need it

src/psse_export.jl
```diff
@@ -813,39 +845,61 @@ function write_to_buffers!(
...
+            end
+            CKT = _psse_quote_string(CKT)
+            B = 0.0
+            RATEA, RATEB, RATEC =
+                with_units_base(exporter.system, PSY.UnitSystem.NATURAL_UNITS) do
+                    _value_or_default(PSY.get_rating(branch), PSSE_DEFAULT),
+                    0.0,
+                    0.0
+                end
+            RATEA = RATEA >= 1e6 ? 0.0 : RATEA
```
GabrielKS, line 876: eliminate magic number

src/psse_export.jl
```diff
@@ -860,7 +914,8 @@ WRITTEN TO SPEC: PSS/E 33.3 POM 5.2.1 Transformer Data
 """
 function _psse_transformer_names(
     transformers::Vector{String},
-    bus_numbers::Vector{Tuple{Int64, Int64}},
+    # bus_numbers::Vector{Tuple{Int64, Int64}},
```
GabrielKS, line 917: ??

src/psse_export.jl
```diff
@@ -875,18 +930,33 @@ function _psse_transformer_names(
...
         end
-        # If both bus numbers are large, that new_name might be too long
         if !_is_valid_psse_name(new_name)
             n = 0
             while !_is_valid_psse_name(new_name) || (new_name in used_names)
-                new_name = "B$(bus_number_mapping[orig_from])-N$n"
+                if length(bus_tuple) == 2
+                    new_name = "B$(bus_number_mapping[bus_tuple[1]])-N$n"
+                else
+                    new_name = "B$(bus_number_mapping[bus_tuple[1]])-N$n"
```
GabrielKS, line 958: same expression in both branches?

src/psse_export.jl
```diff
@@ -933,75 +1022,546 @@ function write_to_buffers!(
...
+
+            if CW == 1
+                R1_2 = PSY.get_r(transformer)
+                X1_2 = PSY.get_x(transformer)
+            elseif CW == 2
+                R1_2 = PSY.get_r(transformer)
+                X1_2 = PSY.get_x(transformer)
+            else
+                R1_2 = PSY.get_r(transformer)
+                X1_2 = PSY.get_x(transformer)
```
GabrielKS, line 1086: same expressions in all branches?

src/psse_export.jl
```diff
@@ -933,75 +1022,546 @@ function write_to_buffers!(
...
+
+            if CW == 1
+                R1_2 = PSY.get_r_12(transformer)
+                X1_2 = PSY.get_x_12(transformer)
+            elseif CW == 2
+                R1_2 = PSY.get_r_12(transformer)
+                X1_2 = PSY.get_x_12(transformer)
+            else
+                R1_2 = PSY.get_r_12(transformer)
+                X1_2 = PSY.get_x_12(transformer)
```
GabrielKS, line 1161: same in all branches?

src/psse_export.jl
```diff
@@ -933,75 +1022,546 @@ function write_to_buffers!(
...
+            singles_to_1 = false,
+        )
+    end
+
+    for (vscline, (from_n, to_n)) in vsc_lines_with_numbers
+        I = md["bus_number_mapping"][from_n]
+        J = md["bus_number_mapping"][to_n]
+        vsc_line_name = string(split(PSY.get_name(vscline), "_")[end])
+        NAME = _psse_quote_string(vsc_line_name)
+        MDC = Int(PSY.get_available(vscline))
```
GabrielKS, line 1417: be consistent — either `get_available ? 1 : 0` everywhere or this everywhere

src/psse_export.jl
```diff
@@ -933,75 +1022,546 @@ function write_to_buffers!(
...
+        g = PSY.get_g(vscline)
+        if g == 0.0
+            RDC = g
+        else
+            RDC =
```
GabrielKS, line 1422: add a comment describing what is going on here

src/psse_export.jl
```diff
@@ -933,75 +1022,546 @@ function write_to_buffers!(
...
+        F3 = PSSE_DEFAULT
+        O4 = PSSE_DEFAULT
+        F4 = PSSE_DEFAULT
+
+        IBUS1 = I
+        TYPE1 = Int(PSY.get_dc_voltage_control_from(vscline))
+        MODE1 = Int(PSY.get_available(vscline))
+        DCSET1 = PSY.get_dc_setpoint_from(vscline)
+        ACSET1 = PSY.get_ac_setpoint_from(vscline)
+        ALOSS1 = 0.0 # TODO: rethink this approach
```
GabrielKS, line 1442: address before merge?

src/psse_export.jl
```diff
@@ -933,75 +1022,546 @@ function write_to_buffers!(
...
+        BLOSS1 =
+            _psse_round_val(
+                PSY.get_proportional_term(
+                    PSY.get_function_data(PSY.get_converter_loss_from(vscline)),
+                ) * 1e3 * PSY.get_base_power(exporter.system))
+        MINLOSS1 = 0.0 # TODO: rethink this approach
+        SMAX1 = PSY.get_rating_from(vscline)
+        SMAX1 = if SMAX1 == 9999.0
```
GabrielKS, line 1450: eliminate magic number

src/psse_export.jl
```diff
@@ -1032,6 +1592,158 @@ function write_to_buffers!(
...
+            push!(N_vals, N)
+            push!(B_vals, imag(B) * PSY.get_base_power(exporter.system))
+        end
+
+        while length(N_vals) < 8
+            push!(N_vals, PSSE_DEFAULT)
+            push!(B_vals, PSSE_DEFAULT)
+        end
+
+        N1 = get(N_vals, 1, PSSE_DEFAULT)
```
GabrielKS, line 1718: use a loop for this?

GabrielKS (changes requested):
Some initial comments. More important than any of that, how is the round trip parsing looking? Is there a branch of PSY we can use such that the round trip stuff in `test_psse_export.jl` works given a system with these types of components? Presumably the existing systems we're testing with do not include all these components — if so, change the systems we're testing with or add additional ones.

src/psse_export.jl
mcllerena: Haven't got there yet, I finished exporting and reading back the system. I'll follow with this and testing the systems.

src/psse_export.jl
mcllerena: For these some of the testing I was doing got caught in the commit. I'll fix it

GabrielKS (changes requested):
The most important things here:

1. I'm not seeing any updates to the test suite? It's important to me that there be unit tests for all the types of components we support.
2. I think your HVDC design will not work with power flow in the loop setpoint updating, see the comment about that.
3. I think your ICT exporting will fail on repeated exports, see the comment about that. Happy to verify if you provide a test system….
4. There are a few places where special cases are triggered by something in a component's name. Beyond the instances of this that are forced by the PSS/E spec ("&", "@", "*"), I don't think we should be doing that, I think we should find another way to keep track of star buses, generators that represent the start and end of HVDC lines, etc.

Otherwise it's mostly small code things.

src/common.jl
```diff
@@ -211,7 +211,8 @@ function make_dc_powerflowdata(
     calculate_loss_factors,
 )
     branch_type = Vector{DataType}(undef, length(branch_lookup))
-    for (ix, b) in enumerate(PNM.get_ac_branches(sys))
+    for (ix, b) in zip(1:length(branch_lookup), PNM.get_ac_branches(sys))
```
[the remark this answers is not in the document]
mcllerena: I've already revert it, this was a small test I was doing, shouldn't have been pushed, mb.

src/psse_export.jl
```diff
@@ -925,85 +1147,588 @@ function write_to_buffers!(
...
+        TYPE1 = get(PSY.get_ext(vscline), "TYPE_FROM", PSSE_DEFAULT)
+        MODE1 = PSY.get_available(vscline) ? 1 : 0
+        DCSET1 = PSY.get_dc_setpoint_from(vscline)
+        ACSET1 = PSY.get_ac_setpoint_from(vscline)
+        ALOSS1 = get(PSY.get_ext(vscline), "ALOSS_FROM", PSSE_DEFAULT)
+        BLOSS1 =
+            _psse_round_val(
+                PSY.get_proportional_term(
+                    PSY.get_function_data(PSY.get_converter_loss_from(vscline)),
+                ) * 1e3 * PSY.get_base_power(exporter.system))
```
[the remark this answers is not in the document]
mcllerena: Regarding N. 1. This is not quite right. The constant term is a sum between Aloss and MINloss, so what I'll be doing in the update is multiply the proportional terminal with the DC current to get a positive Aloss and then substract the constant term to get a positive MINloss. I'll very close results to the ones in the original file, otherwise I do not see how can I extract those values.

src/psse_export.jl
```diff
@@ -705,14 +764,76 @@ function write_to_buffers!(
...
+            if !haskey(generators_by_bus, bus_num)
+                generators_by_bus[bus_num] = []
+            end
+            push!(generators_by_bus[bus_num], (PSY.get_name(generator), i))
+        end
+
+        # Create mapping with sequential numbering per bus
+        mapping = Dict{Tuple{Int, String}, String}()
+        for (bus_num, gens_on_bus) in generators_by_bus
+            if length(gens_on_bus) == 1
```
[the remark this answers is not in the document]
mcllerena: This updates the new created gens (bus number, ID) from the GenericHVDC. Updates the ID accordingly in case there's more than 1 generator in the bus after adding the new gens.

src/psse_export.jl
```diff
@@ -449,6 +461,10 @@ function write_to_buffers!(
     bus_name_mapping = md["bus_name_mapping"]
 
     for bus in buses
+        bus_name = PSY.get_name(bus)
+        if startswith(strip(bus_name), "starbus")
```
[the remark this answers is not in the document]
jd-lara: we need to read it from the 3WT transformers before hand

src/psse_export.jl
```diff
@@ -705,14 +764,76 @@ function write_to_buffers!(
...
+        if !isempty(hvdc_lines)
+            @warn "Found $(length(hvdc_lines)) TwoTerminalGenericHVDCLine components. These will be exported as generators at each end of the DC line."
+
+            for hvdc_line in hvdc_lines
+                from_bus = PSY.get_from(PSY.get_arc(hvdc_line))
+                to_bus = PSY.get_to(PSY.get_arc(hvdc_line))
+
+                push!(
+                    temp_gens,
+                    _make_gens_from_hvdc(
```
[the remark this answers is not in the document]
jd-lara: we need to of with option 2 here. is the only viable alternative really

src/psse_export.jl
```diff
@@ -796,12 +1006,16 @@ function write_to_buffers!(
...
             collect(
-                PSY.get_components(Union{PSY.Line, PSY.MonitoredLine}, exporter.system),
+                PSY.get_components(
+                    Union{PSY.Line, PSY.MonitoredLine, PSY.DiscreteControlledACBranch},
+                    exporter.system,
+                ),
             );
             by = branch_to_bus_numbers,
         )
         [(branch, branch_to_bus_numbers(branch)) for branch in branches]
```
jd-lara, line 1016: this code is pretty crazy, please implement it in a function with adequate comments.

PR #222, opened 2025-08-05
PNM interface changes

luke-kiernan:
Fix things for PNM interface changes. The exporter tests error--the newly added `winding_group_number` fields don't match--but I'll leave that for @rbolgaryn to fix. Otherwise, things are back to where they were a week ago: 20-30 test failures, all due to low-priority issues (reactive power distribution, issue #201).

edit: I also moved the loss factor and distributed slack tests into their own files. If vanilla power flows aren't working, those tests bury those errors behind a wall of similar messages. I don't use `failfast` because it stops the whole testing process, not just the current test set.


rbolgaryn:
Looks good!

src/post_processing.jl
```diff
@@ -548,8 +548,8 @@ end
 
 # returns list of branches names and buses numbers: PTDF and virtual PTDF case
 function _get_branches_buses(data::Union{PTDFPowerFlowData, vPTDFPowerFlowData})
-    return PNM.get_branch_ax(data.power_network_matrix),
-    PNM.get_bus_ax(data.power_network_matrix)
+    return PNM.get_arc_axis(data.power_network_matrix),
+    PNM.get_bus_axis(data.power_network_matrix)
```
m-bossart, line 552: This returns a list of arc tuples, not the branches names so the function name and comment should be updated

luke-kiernan: Fixed. @Jose I know you've said that `branch_lookup` should have strings as keys, not tuples. I'll fix that in a separate PR: get things working first, then make design changes.

luke-kiernan: Hmm. PNM's data structures all have tuples as keys, though. How about 2 separate lookups, `arc_lookup` (keys are tuples) and `branch_lookup` (keys are strings)? Is that ok? I could make it work with a single lookup, but the side effects would be

1. I can only use the arc/branch-related network reduction data in places where I have access to the `PSY.System`. I have the names of the branches: the network reduction data requires `PSY.Branch` objects or tuples.
2. I'd have to name the reduced arcs in such a way that I can retrieve the arc. E.g. as `from-to`. Otherwise, after figuring out the flows, I have no way of knowing "flow in what?" So I'd be dealing in terms of arcs for the reduced branches anyway, just indirectly.

jd-lara: > Hmm. PNM's data structures all have tuples as keys, though. How about 2 separate lookups, `arc_lookup` (keys are tuples) and `branch_lookup` (keys are strings)? Is that ok? I could make it work with a single lookup with keys as strings, but the side effects would be
> 
> 1. I can only use the arc/branch-related network reduction data in places where I have access to the `PSY.System`. I have the names of the branches: the network reduction data requires `PSY.Branch` objects or tuples.
> 2. I'd have to name the reduced arcs in such a way that I can retrieve the arc. E.g. as `from-to`. Otherwise, after figuring out the flows, I have no way of knowing "flow in what?" So I'd be dealing in terms of arcs for the reduced branches anyway, just indirectly.

you need to exploit the `NetworkReductionData` object in PNM to do this. The map you are talking about is already there.

luke-kiernan: > The map you are talking about is already there.

I don't follow. Which map do you have in mind?

If `branch_lookup` has type `Dict{String, Int}`, that only gives me branch/arc name to index. All the `branch`-related maps in `NetworkReductionData` that I see have key type `Tuple{Int, Int}` or `PSY.Branch`.

src/common.jl
```diff
@@ -230,7 +230,7 @@ function make_dc_powerflowdata(
     for (ix, b) in enumerate(PNM.get_ac_branches(sys))
         branch_type[ix] = typeof(b)
     end
-    bus_reactivepower_bounds = Vector{Vector{Float64}}(undef, n_buses)
+    # bus_reactivepower_bounds = Vector{Vector{Float64}}(undef, n_buses)
```
jd-lara, line 233: why is this commented out?

jd-lara (approved):
one minor comment before merge

jd-lara (approved):

PR #224, opened 2025-08-05
Transformer2W's `primary_shunt` is complex

luke-kiernan:
Update `post_processing.jl` for the shunt-related changes to `Line` (new `g` field) and `Transformer2W` (`primary_shunt` is now complex).


rbolgaryn (changes requested):
TapTransformer also has primary_shunt

PR #227, opened 2025-08-25
Allow user to specify `x0`

luke-kiernan:
If the user changes `OVERRIDE_x0` in `definitions.jl` to `true` and passes `:x0 = v` to `solve_powerflow!`, use that `v` as the starting point. [Moreover, it uses `v` as-is: even if the residual is large and `robust_power_flow` is `true`, it doesn't do a DC power flow to improve the initial guess.] This way, we can test performance and convergence robustness on different `x0`'s, instead of always using the set points from the system objects.


jd-lara (commented):
I am not crazy about this approach. There needs to be a way to do this that doesn't require redefining a constant and adding this brute force approach. @daniel-thom any suggestions?

luke-kiernan:
> I am not crazy about this approach. There needs to be a way to do this that doesn't require redefining a constant and adding this brute force approach.

What do you mean by "adding this brute force approach?"

I can use a keyword argument instead of a global constant: I did things this way to keep it as a developer-oriented feature. Also, if it's undesireable to have the `x0` kwarg bypass initial guess improvements, then I can change that too. But it sounds like that's not what you have in mind.

PR #228, opened 2025-08-26
Robust Homotopy performance improvements

luke-kiernan:
Further performance improvements: these drop the runtime of the Robust Homotopy method from 6 seconds to 2.4 seconds. The vast majority of the improvement (6s -> 2.6s) is from switching to `LineSearches.jl`'s back tracking line search.


src/RobustHomotopy/homotopy_hessian.jl
```diff
@@ -45,6 +45,7 @@ function F_value(hess::HomotopyHessian, t_k::Float64, x::Vector{Float64}, time_s
     return F_value
 end
 
+# slightly confusing that I have the field grad, and the argument grad.
```
jd-lara, line 48: what's the point of this comment?

src/RobustHomotopy/robust_homotopy_method.jl
```diff
@@ -100,10 +100,14 @@ function _second_order_newton_step(homHess::HomotopyHessian,
     solve!(hSolver, δ)
     δ .*= -1
 
-    # PERF: the line search is taking up 80%+ of the time in _second_order_newton_step
-    # evidently I need a better (or better optimized) line search.
-    # (The line search is non-allocating, except for the @warn message.)
-    α_star = line_search(x, time_step, homHess, t_k, δ)
+    # Create objective function
+    ϕ = α -> F_value(homHess, t_k, x + α * δ, time_step)
```
jd-lara, line 104: The problem with this is that it creates the function at run time so the code can't really be compiled.

PR #234, opened 2025-09-04
bus voltages and power flows in series, parallel branches

luke-kiernan:
Make it so `solve_powerflow!(pf, sys)` sets the bus voltages and branch flows in the system objects removed in a degree 2 reduction. Also set the flows in parallel branches. I've tested this on the 2k bus system: the recovered voltages/flows match to a good degree of precision.

I added a `skip_redistribution` field to `ACPowerFlow`, as a work-around for some bugs in the active/reactive power redistribution code in `post_processing.jl`. This can be removed once that code is working again: fixing it is the next item on my to do list.

One design decision that may merit discussion: I realized that `post_processing.jl` was setting the system's bus angle (PV) or angle/voltage (PQ) based on the system bus types, not `data.bus_type`. If `correct_bustypes` is `true`, that's a problem. Here's the design choice: if `correct_bustypes` is `true`, then the update-system `solve_powerflow!(pf, sys)` now changes PV buses without any generation to PQ and prints a warning.

edit: relies on PNM's [PR #182](https://github.com/NREL-Sienna/PowerNetworkMatrices.jl/pull/182)


src/post_processing.jl
```diff
@@ -340,32 +340,85 @@ function _reactive_power_redistribution_pv(
...
-        PSY.set_reactive_power_flow_primary!(trf, imag(flow))
-    elseif winding == 2
-        PSY.set_active_power_flow_secondary!(trf, real(flow))
-        PSY.set_reactive_power_flow_secondary!(trf, imag(flow))
-    elseif winding == 3
-        PSY.set_active_power_flow_tertiary!(trf, real(flow))
-        PSY.set_reactive_power_flow_tertiary!(trf, imag(flow))
-    else
-        error("Invalid winding number: $winding")
+function _set_series_voltages_and_flows!(
```
rbolgaryn, line 343: The variable names here should be written out

src/post_processing.jl
```diff
@@ -392,45 +445,45 @@ Updates system voltages and powers with power flow results
...
+    # FIXME once redistribution is working again, could remove skip_redistribution.
+    if !pf.skip_redistribution
+        source_buses = can_be_PV(sys)
+        for (bus_number, reduced_buses) in PNM.get_bus_reduction_map(nrd)
+            if length(reduced_buses) == 0
+                # no reduction.
+                bus_name = temp_bus_map[bus_number]
+                bus = PSY.get_component(PSY.ACBus, sys, bus_name)
+                system_bustype = PSY.get_bustype(bus)
+                ix = get_bus_lookup(data)[bus_number]
```
rbolgaryn, line 469: get_bus_lookup should be outside of the loop

src/post_processing.jl
```diff
@@ -392,45 +445,45 @@ Updates system voltages and powers with power flow results
...
+                data_bustype = data.bus_type[ix, time_step]
+                (system_bustype == data_bustype) && continue
+                # assumption: no sources => bus type must've been changed during initialization
+                # due to the correct_bustypes option, not due to reactive power limits.
+                !(bus_number in source_buses) && continue
```
rbolgaryn, line 474: nitpick: can be ```(bus_number in source_buses) || continue```

src/post_processing.jl
```diff
@@ -446,24 +499,38 @@ function write_powerflow_solution!(
             bus_name = temp_bus_map[bus_number]
             bus = PSY.get_component(PSY.ACBus, sys, bus_name)
             ix = get_bus_lookup(data)[bus_number]
-            if bus.bustype == PSY.ACBusTypes.REF
+            bustype = data.bus_type[ix, time_step] # may not be the same as bus.bustype!
```
rbolgaryn, line 502: This for-loop has some code duplication. Is it because the above for loop will be removed?

src/powersystems_utils.jl
```diff
@@ -0,0 +1,53 @@
...
+
+function set_voltage!(bus::PSY.ACBus, V::Complex)
+    PSY.set_magnitude!(bus, abs(V))
+    PSY.set_angle!(bus, angle(V))
+    return
+end
+
+get_complex_voltage(bus::PSY.ACBus) = PSY.get_magnitude(bus) * exp(1im * PSY.get_angle(bus))
+
+function calculate_segment_flow!(
```
rbolgaryn, line 32: are we only ever interested in the power flow at 'from' buses? Is this because PSY only has one entry for power flow?

src/post_processing.jl
```diff
@@ -392,45 +445,45 @@ Updates system voltages and powers with power flow results
...
+        for (bus_number, reduced_buses) in PNM.get_bus_reduction_map(nrd)
+            if length(reduced_buses) == 0
+                # no reduction.
+                bus_name = temp_bus_map[bus_number]
+                bus = PSY.get_component(PSY.ACBus, sys, bus_name)
+                system_bustype = PSY.get_bustype(bus)
+                ix = get_bus_lookup(data)[bus_number]
+                data_bustype = data.bus_type[ix, time_step]
+                (system_bustype == data_bustype) && continue
+                # assumption: no sources => bus type must've been changed during initialization
```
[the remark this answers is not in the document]
jd-lara: Why are we changing the system, this is always kinda problematic

PR #235, opened 2025-09-05
PV-PQ warning when correct_bustypes = true

luke-kiernan:
Move PV-PQ validation stuff from PSY to PF. See issue [#1525](https://github.com/NREL-Sienna/PowerSystems.jl/issues/1525#issuecomment-3254502219) in PSY.


rbolgaryn (approved):
is good

PR #242, opened 2025-09-15
add lcc hvdc to PF

rbolgaryn:


src/PowerFlowData.jl
```diff
@@ -15,6 +15,34 @@ abstract type SystemPowerFlowContainer <: PowerFlowContainer end
...
+
+    dc_line_resistance::Vector{Float64}
+
+    rectifier_tap::Matrix{Float64}
+    inverter_tap::Matrix{Float64}
+
+    rectifier_delay_angle::Matrix{Float64}     # α
+    inverter_extinction_angle::Matrix{Float64} # γ
+
+    rectifier_phi::Matrix{Float64}
```
GabrielKS, line 33: This is okay for me, I'll just note that sometimes we use special characters (search PSY codebase for `α`, `γ`, see also PF `robust_homotopy_method.jl`), we should probably be consistent about whether we use special characters or their names.

src/PowerFlowData.jl
```diff
@@ -130,6 +158,7 @@ struct PowerFlowData{
     calculate_loss_factors::Bool
     voltage_stability_factors::Union{Matrix{Float64}, Nothing}
     calculate_voltage_stability_factors::Bool
+    lcc::Union{LCCParameters, Nothing}
```
GabrielKS, line 161: @luke-kiernan I'll defer to you on the wiseness of this union type

src/PowerFlowData.jl
```diff
@@ -256,6 +285,95 @@ get_calculate_loss_factors(pfd::PowerFlowData) = pfd.calculate_loss_factors
...
+    lcc_rectifier_bus = zeros(Int, n_lccs)
+    lcc_inverter_bus = zeros(Int, n_lccs)
+    lcc_rectifier_transformer_reactance = zeros(Float64, n_lccs)
+    lcc_inverter_transformer_reactance = zeros(Float64, n_lccs)
+    lcc_rectifier_min_alpha = zeros(Float64, n_lccs)
+    lcc_inverter_min_gamma = zeros(Float64, n_lccs)
+
+    lccs = PSY.get_components(PSY.get_available, PSY.TwoTerminalLCCLine, sys)
+    _get_lcc_bus_ix =
+        (bus) -> _get_bus_ix(bus_lookup, reverse_bus_search_map, PSY.get_number(bus))
```
GabrielKS, line 334: I admire the attempt to make more elegant but given that it's only used twice I'd say we should just call `_get_bus_ix` directly rather than incurring the overhead of this dynamically defined function

src/ac_power_flow_jacobian.jl
```diff
@@ -198,6 +198,82 @@ function _create_jacobian_matrix_structure_bus!(rows::Vector{Int32},
...
+        push!(columns, num_buses * 2 + (i - 1) * 4 + 4)
+        push!(values, 0.0)
+
+        push!(rows, num_buses * 2 + (i - 1) * 4 + 3)
+        push!(columns, num_buses * 2 + (i - 1) * 4 + 3)
+        push!(values, 1.0)
+
+        push!(rows, num_buses * 2 + (i - 1) * 4 + 4)
+        push!(columns, num_buses * 2 + (i - 1) * 4 + 4)
+        push!(values, 1.0)
```
GabrielKS, line 272: Is it possible to define a helper function to streamline some of this implementation? It all feels a little magical for me, maybe some comments would also be useful

src/ac_power_flow_jacobian.jl
```diff
@@ -380,6 +458,105 @@ function _set_entries_for_neighbor(Jv::SparseArrays.SparseMatrixCSC{Float64, Int
...
+        Jv[num_buses * 2 + (i - 1) * 4 + 2, num_buses * 2 + (i - 1) * 4 + 2] =
+            data.bus_magnitude[tb, time_step] * sqrt(6) / π *
+            data.lcc.inverter_i_dc[i, time_step] *
+            cos(data.lcc.inverter_extinction_angle[i, time_step])
+        Jv[num_buses * 2 + (i - 1) * 4 + 2, num_buses * 2 + (i - 1) * 4 + 3] =
+            -data.bus_magnitude[fb, time_step] * data.lcc.rectifier_tap[i, time_step] *
+            sqrt(6) / π * data.lcc.rectifier_i_dc[i, time_step] *
+            cos(data.lcc.rectifier_delay_angle[i, time_step]) *
+            tan(data.lcc.rectifier_delay_angle[i, time_step])
+        Jv[num_buses * 2 + (i - 1) * 4 + 2, num_buses * 2 + (i - 1) * 4 + 4] =
```
GabrielKS, line 551: I'm a little worried that so much array index manipulation will make it hard for someone else to come in and work with this code if something breaks in a few years — like, if there is a typo and one of the `4`s is a `3`, will they be able to find that? There is a joke that complicated regular expressions are a "write-only language" because once they are written no one can ever understand them again. I'm not really sure how to do it better, though. @luke-kiernan any ideas?

GabrielKS (commented):
A few initial code quality things. Needs tests of course.

src/ac_power_flow_jacobian.jl
luke-kiernan: The products `data.bus_magnitude[fb, time_step] * data.lcc.inverter_i_dc[i, time_step]` and `data.bus_magnitude[tb, time_step] * data.lcc.rectifier_i_dc[i, time_step]` occur very frequently. Compute those once, assign them some variable, then use that variable.

The above 4 expressions occur in 2 pairs of 2: where the first one has `fb` and `rectifier` on the right side, the second has `tb` and `inverter`. Does that pattern hold more broadly?

src/PowerFlowData.jl
luke-kiernan: Yeah it's probably fine. I dislike how messy the `PowerFlowData` struct is becoming, but until we do a full overhaul, I don't see good alternatives.

src/PowerFlowData.jl
```diff
@@ -15,6 +15,34 @@ abstract type SystemPowerFlowContainer <: PowerFlowContainer end
 
 get_system(container::SystemPowerFlowContainer) = container.system
 
+struct LCCParameters
```
luke-kiernan, line 18: Enough of these fields occur in `rectifier_foo` and `inverter_foo` pairs that I'd suggest introducing an inner struct and storing 2 of them, one for rectifier and one for inverter. I suspect you could write helper functions on that inner struct that work for both cases, cutting down on code duplication.

luke-kiernan (commented):
2 main concerns:

1. The math on the right side is very difficult to follow. I've made some comments about how that might be improved. The biggest one: introduce an inner struct and write helper functions on that struct that work for both `rectifier` and `inverter` cases.
2. The indexing expressions are also hard to follow. Currently, we have variables grouped by bus: 2 inputs, 2 outputs per bus. This PR adds 4 new inputs and outputs on the end, breaking that regularity. How often will systems have LCCs? If the answer is "not very often," I'd wonder about storing the Jacobian in separate 4 blocks: top-left with the `2n`x`2n` entries for the buses, other 3 involving the LCCs. Similarly, `x` and `F` would split into 2 vectors: one for the bus entries, one for the LCC entries.

luke-kiernan (commented):
Looks good, for the most part. A few big-picture concerns:

1. How does this interact with network reduction stuff? Can LCC lines be part of a series or radial reduction? How about parallel lines? [I think not on that last one.] If "yes," then we should add a test case that calls `solve_powerflow!(pf, sys)` on a system where that happens; if "no," then add some error or warning against it. Doesn't need to happen now: would be sufficient to open an issue.
2. Accuracy of the Jacobian expressions. Could you add a test case where you compare a numerical derivative against the Jacobian? i.e. pick some point `x` and direction vector `u`, then verify that `[F(x + t u)-F(x)]/t` approaches `J(x) u` as t goes to 0.
3. Without LCCs, the ybus entries don't change throughtout the power flow, placing the struct firmly under the domain of PNM. With LCCs, the ybus matrix now depends on the state vector. This complicates the relationship between PNM and PF...

There's a couple "not implemented" errors and the like: please resolve those before you merge.

Also, how are LCC lines handled in DC power flows?

luke-kiernan (changes requested):
Could you make the Jacobian test more in line with PR #246? Checking that the error goes to 0 linearly isn't a meaningful criteria: that also happens if you replace J with any other matrix...

src/common.jl
```diff
@@ -655,3 +659,113 @@ wdot(wx::Vector{Float64}, x::Vector{Float64}, wy::Vector{Float64}, y::Vector{Flo
...
+    I_dc::Float64,
+    x_t::Float64,
+    Vm::Float64,
+    ϕ::Float64,
+    α::Float64,
+)::Float64
+    return Vm * t * sqrt(6) / π * I_dc * cos(ϕ) * sin(α) / sqrt(1 - cos(ϕ)^2)
+end
+
+function _update_ybus_lcc!(ybus_facts, data, time_step)
```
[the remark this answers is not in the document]
luke-kiernan: Never mind. I realized that `ybus_facts` is used in one place in `power_flow_residual.jl`

PR #245, opened 2025-09-18
refactor power flow data constructor

luke-kiernan:
Overhaul of the `PowerFlowData` constructor. Changes:

1. Instead of storing `{bus/arc}_{lookup/axis}` as a field, just patch the calls through to the PNM object. Same for `valid_ix`. This means we don't need to calculate them in the bodies of the constructors.
2. Instead of passing arguments all the way down the call stack and creating the `PowerFlowData` object at the very bottom, we create a `PowerFlowData` object first and initialize the fields from the system second. This way, our functions stay short, require fewer arguments, and run the same for the different power flow types (with the differences quietly handled by multiple dispatch).

There is more I'd like to do along these lines, but I wanted to take things one step at a time. Some notes to the reviewer that may be helpful:

- `initialize_powerflow_data!` is mostly a copy-paste of the guts of `make_powerflowdata`
- `make_bus_slack_participation_factors` is basically unchanged, just happens in-place.
- the things that used to happen in the individual `PowerFlowData` constructors are either unnecessary (`{bus/arc}_{lookup/axis}`, `valid_ix`) or in `make_powerflow_data` (`timestep_map`).

### DIFF
--- a/src/PowerFlowData.jl
+++ b/src/PowerFlowData.jl
@@ -23,12 +23,6 @@ In the below descriptions, "number of buses" should be understood as "number of 
 after the network reduction." Similarly, we use "arcs" instead of "branches" to distinguish 
 between network elements (post-reduction) and system objects (pre-reduction).
 # Arguments:
-- `bus_lookup::Dict{Int, Int}`:
-        dictionary linking the system's bus number with the rows of either
-        "power_network_matrix" or "aux_network_matrix".
-- `arc_lookup::Dict{Tuple{Int, Int}, Int}`:
-        dictionary linking the arc name with the column name of either the
-        "power_network_matrix" or "aux_network_matrix".
 - `bus_activepower_injection::Matrix{Float64}`:
         "(b, t)" matrix containing the bus active power injection. b: number of
         buses, t: number of time period.
@@ -83,8 +77,6 @@ between network elements (post-reduction) and system objects (pre-reduction).
 - `timestep_map::Dict{Int, S}`:
         dictionary mapping the number of the time periods (corresponding to the
         column number of the previously mentioned matrices) and their names.
-- `valid_ix::Vector{Int}`:
-        vector containing the indices of not slack buses
 - `power_network_matrix::M`:
         matrix used for the evaluation of either the power flows or bus angles,
         depending on the method considered.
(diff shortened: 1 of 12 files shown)


src/initialize_powerflow_data.jl
```diff
@@ -0,0 +1,107 @@
...
+        sys,
+        correct_bustypes,
+    )
+    for i in 1:size(data.bus_type, 2)
+        data.bus_type[:, i] .= bus_type
+    end
+    data.bus_angles[:, 1] .= bus_angles
+    data.bus_magnitude[:, 1] .= bus_magnitude
+
+    # active power injections, withdrawals
```
m-bossart, line 39: active and reactive power injections

m-bossart (approved):
This looks ok to me, one typo in a comment

src/PowerFlowData.jl
```diff
@@ -512,49 +452,21 @@ function PowerFlowData(
...
-    valid_ix = setdiff(1:length(bus_lookup), PNM.get_ref_bus_position(power_network_matrix))
-    converged = fill(false, time_steps)
-    loss_factors = nothing
-    calculate_loss_factors = false
-    generator_slack_participation_factors = nothing
-    voltage_stability_factors = nothing
-    calculate_voltage_stability_factors = false
-    return make_dc_powerflowdata(
+    return make_and_initialize_powerflow_data(
+        PTDFDCPowerFlow(),
```
luke-kiernan, line 463: I could just directly pass the `pf` we were given as input, but the difference doesn't actually matter because we don't use the `exporter` field anywhere here in PF, only in PSI.

src/PowerFlowData.jl
```diff
@@ -309,6 +298,28 @@ function network_reduction_message(
...
+    pf::PowerFlowEvaluationModel,
+    sys::PSY.System,
+    power_network_matrix::M,
+    aux_network_matrix::N;
+    time_steps::Int = 1,
+    timestep_names::Vector{String} = String[],
+    neighbors = Vector{Set{Int}}(),
+    correct_bustypes::Bool = false,
+) where {M <: PNM.PowerNetworkMatrix, N <: Union{PNM.PowerNetworkMatrix, Nothing}}
+    data = make_powerflow_data(
```
luke-kiernan, line 311: Could just make this another constructor, `PowerFlowData(...)`, instead of calling it `make_powerflow_data(...)`.

src/common.jl
```diff
@@ -355,53 +317,65 @@ function make_bus_slack_participation_factors(
...
             reverse_bus_search_map,
             time_steps,
             n_buses,
             bus_type,
         )
+        return
     end
 
-    if length(generator_slack_participation_factors) < time_steps
+    if length(generator_slack_participation_factors_input) != time_steps
```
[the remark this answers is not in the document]
luke-kiernan: I'll add a warning if it's longer. At least that eliminates the above concern.

Issue #184, opened 2025-07-02
Use multiple dispatch in `solve_dc_powerflow.jl`


luke-kiernan:
Currently, we have 3 single-period DC power flow solvers, and 3 multi-period DC power flow solvers. Yet their bodies are identical, except for one type choice (vPTDF vs PTDF vs ABA): this is a textbook use case for multiple dispatch. Collapse the three into one and combine their docstrings.

PR #246, opened 2025-09-25
add tests for jacobian, for arc types in network reductions

luke-kiernan:
Add some tests: one set verifying that `J` matches a numerical derivative, another akin to [PR #184](https://github.com/NREL-Sienna/PowerNetworkMatrices.jl/pull/184) in PNM, testing branch types in network reductions.

### DIFF
--- a/test/test_arc_types_and_reductions.jl
+++ b/test/test_arc_types_and_reductions.jl
@@ -11,18 +11,17 @@ function test_all_powerflow_types(
         correct_bustypes = true,
     )
     solve_powerflow!(data; pf = pf) # should run without errors.
-    @test data isa Any
+    @test !isempty(data.arc_activepower_flow_from_to)
     solve_powerflow!(
         pf,
         sys;
         network_reductions = deepcopy(network_reductions),
         correct_bustypes = true,
     )
-    for pf in [PF.DCPowerFlow(), PF.PTDFDCPowerFlow(), PF.vPTDFDCPowerFlow()]
-        data = PF.PowerFlowData(pf, sys; network_reductions = deepcopy(network_reductions))
-        solve_powerflow!(data) # should run without errors.
-        @test data isa Any
-    end
+    pf = PF.DCPowerFlow()
+    data = PF.PowerFlowData(pf, sys; network_reductions = deepcopy(network_reductions))
+    solve_powerflow!(data) # should run without errors.
+    @test !isempty(data.arc_activepower_flow_from_to)
 end
 
 @testset "radial reduction with 3WT/parallel lines" begin
--- a/test/test_jacobian.jl
+++ b/test/test_jacobian.jl
@@ -1,5 +1,5 @@
 @testset "Jacobian Verification" begin
-    sys = build_system(MatpowerTestSystems, "matpower_ACTIVSg2000_sys")
+    sys = PSB.build_system(PSB.PSITestSystems, "c_sys14")
     pf = PF.ACPowerFlow{NewtonRaphsonACPowerFlow}()
     data = PF.PowerFlowData(pf, sys; correct_bustypes = true)
     time_step = 1
@@ -24,6 +24,10 @@
                 push!(outputValues, deepcopy(residual.Rv))
             end
             ΔF = outputValues[2] .- outputValues[1]
+            floating_point_issues = all(isapprox.(ΔF, 0.0; atol = eps(Float32)))
+            if floating_point_issues
+                break
+            end
             ∂F_∂u_numerical = ΔF ./ Δx_mag
             ∂F_∂u_symbolic = J_x0 * u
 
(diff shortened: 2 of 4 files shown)


test/test_arc_types_and_reductions.jl
```diff
@@ -0,0 +1,38 @@
...
+    # AC powerflow has different syntax
+    pf = ACPowerFlow()
+    data = PF.PowerFlowData(
+        pf,
+        sys;
+        network_reductions = deepcopy(network_reductions),
+        correct_bustypes = true,
+    )
+    solve_powerflow!(data; pf = pf) # should run without errors.
+    @test data isa Any
```
rbolgaryn, line 14: What is the purpose of this check?

test/test_jacobian.jl
```diff
@@ -0,0 +1,38 @@
...
+    data = PF.PowerFlowData(pf, sys; correct_bustypes = true)
+    time_step = 1
+    residual = PF.ACPowerFlowResidual(data, time_step)
+    J = PF.ACPowerFlowJacobian(data, time_step)
+    n = 2 * length(collect(get_components(ACBus, sys)))
+    x0 = PF.calculate_x0(data, time_step)
+    residual(x0, time_step)
+    J(time_step)
+    J_x0 = deepcopy(J.Jv)
+    Δx_start, Δx_stop = 1e-3, 1e-6
```
rbolgaryn, line 13: why not just do 1e-6 and spare the internal loop?

test/test_jacobian.jl
```diff
@@ -0,0 +1,38 @@
+@testset "Jacobian Verification" begin
+    sys = build_system(MatpowerTestSystems, "matpower_ACTIVSg2000_sys")
```
rbolgaryn, line 2: would it be sufficient to use a smaller grid model for this test?

test/test_arc_types_and_reductions.jl
jd-lara: Same, why does this matter?

test/test_arc_types_and_reductions.jl
```diff
@@ -0,0 +1,38 @@
+function test_all_powerflow_types(
+    sys::System,
+    network_reductions::Vector{PNM.NetworkReduction},
+)
+    # AC powerflow has different syntax
+    pf = ACPowerFlow()
```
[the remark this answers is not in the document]
luke-kiernan: Not necessary: the iterative `solve_powerflow!` process doesn't depend on the branch types. The post processing (and maybe the constructor) do depend on the branch types in spots, but that's the same for all the different types of powerflows.

test/test_jacobian.jl
luke-kiernan: Too big of a step and the test will fail; too small and floating point error ruins becomes a problem. That was the idea, anyway...but now that I think about it, if `Δx` is too small, then `F(x + Δx) - F(x)` will be the zero vector. So maybe I should add `!isapprox(F(x + Δx), F(x))`, too.

test/test_arc_types_and_reductions.jl
```diff
@@ -0,0 +1,38 @@
...
+    solve_powerflow!(
+        pf,
+        sys;
+        network_reductions = deepcopy(network_reductions),
+        correct_bustypes = true,
+    )
+    for pf in [PF.DCPowerFlow(), PF.PTDFDCPowerFlow(), PF.vPTDFDCPowerFlow()]
+        data = PF.PowerFlowData(pf, sys; network_reductions = deepcopy(network_reductions))
+        solve_powerflow!(data) # should run without errors.
+        @test data isa Any
```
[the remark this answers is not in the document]
luke-kiernan: Sure, I can change it to `@test !isempty(data.branch_activepower_flow_from_to)` or something like that.

test/test_arc_types_and_reductions.jl
```diff
@@ -0,0 +1,38 @@
...
+    )
+    solve_powerflow!(data; pf = pf) # should run without errors.
+    @test data isa Any
+    solve_powerflow!(
+        pf,
+        sys;
+        network_reductions = deepcopy(network_reductions),
+        correct_bustypes = true,
+    )
+    for pf in [PF.DCPowerFlow(), PF.PTDFDCPowerFlow(), PF.vPTDFDCPowerFlow()]
```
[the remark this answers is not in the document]
luke-kiernan: Oh, I can just use `subtypes(PF.AbstractDCPowerFlow)` instead.

test/test_arc_types_and_reductions.jl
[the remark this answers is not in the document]
luke-kiernan: Oh, I guess by the same logic testing the different DC power flow types are also unnecessary. I'll switch to just testing one of them.

PR #254, opened 2025-11-03
make compatible with pnm v0.14

luke-kiernan:
At this point, 9 tests are failing. I've marked all of them with `FIXME`. Here's the breakdown:

- `"PSSE Exporter with pti_case24_sys.raw, v33"`[6 failures]: here, both a `@test_log` and a `@test` are failing, which seems to say something's wrong. Might be the exporter, or might be the `compare_systems_loosely` helper function we've using in the test case.
- `"AC Test 240 Case PSS/e results"` [2 failures]: triggers an assertion in PNM at [Ybus.jl:401](https://github.com/NREL-Sienna/PowerNetworkMatrices.jl/blob/main/src/Ybus.jl#L401), indicating that not all branches among a `BranchesParallel` container are oriented in the same direction. @m-bossart could you please look into that?
- `" test_loss_factors_case_14"` with L-M solver [1 failure]: I'm pretty confident that the initial condition just happens to perform poorly with that method. Will confirm once the above 2 are fixed.


jd-lara:
the 240 bus test failure is harder to track. I'll comment it for now

PR #260, opened 2025-11-05
Consistently use `SYSTEM_BASE`

### LINKED ISSUE #256: Line flow results and `UnitSystem` settings
When returning the result dataframes for power flows, we say "Powers are exported in MW/MVAr." Yet the following shows that our calculated flows depend on the `UnitSystem` choice:
```julia
using PowerSystems, PowerFlows, PowerSystemCaseBuilder
const PSY = PowerSystems
function power_flow_with_units(sys::PSY.System, units::PSY.UnitSystem)
    with_units_base(sys, units) do
        results = solve_powerflow(DCPowerFlow(), sys; correct_bustypes = true)
        first_line_flow = results["1"]["flow_results"][1, :]
        return (first_line_flow[:line_name], first_line_flow[:P_from_to])
    end
end
sys = build_system(PSITestSystems, "c_sys14"; add_forecasts = false)
line_name, flow_natural = power_flow_with_units(sys,PSY.UnitSystem.NATURAL_UNITS)
line_name2, flow_system = power_flow_with_units(sys,PSY.UnitSystem.SYSTEM_BASE)
@assert line_name == line_name2
@assert flow_natural == flow_system "flow results depend on units"
```
Looks like we're implicitly assuming that `solve_powerflow` is called in a context where we're working in the system base.

luke-kiernan:
Resolve #256 by adding `with_unit_base`. Also add tests that ensure our reported line flows don't depend on the unit system.

With these changes, I'm getting a few new failures in `"System <-> PowerFlowData round trip"`: might be an actual bug or might be due to PSY issue [#1575](https://github.com/NREL-Sienna/PowerSystems.jl/issues/1575). Will revisit once that issue is resolved.

### DIFF
--- a/src/PowerFlowData.jl
+++ b/src/PowerFlowData.jl
@@ -399,16 +399,18 @@ function make_and_initialize_powerflow_data(
     neighbors = Vector{Set{Int}}(),
     correct_bustypes::Bool = false,
 ) where {M <: PNM.PowerNetworkMatrix, N <: Union{PNM.PowerNetworkMatrix, Nothing}}
-    data = PowerFlowData(
-        pf,
-        power_network_matrix,
-        aux_network_matrix,
-        time_steps;
-        timestep_names = timestep_names,
-        neighbors = neighbors,
-    )
-    initialize_powerflow_data!(data, pf, sys; correct_bustypes = correct_bustypes)
-    return data
+    with_units_base(sys, "SYSTEM_BASE") do
+        data = PowerFlowData(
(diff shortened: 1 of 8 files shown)


GabrielKS (commented):
Seems like a good approach, though not ready for review if tests are still failing

luke-kiernan:
Solved: a helper function in `test/common.jl` was calling the power redistribution functions with the system in `DEVICE_BASE` units.

PR #262, opened 2025-11-09
Jacobian diagnostics

jd-lara:
I updated the code to be more about Jacobian diagnostics. It doesn't run inside of a loop yet but we can use it for detection of bad cases.


luke-kiernan (changes requested):
Gave it a quick once-over.

Math theory question: why are we looking at the eigenvalues of the Jacobian? Negative curvature is a problem, but curvature is 2nd order information. The Jacobian contains only 1st order information. Now, there are ways to approximate 2nd order info with 1st order info, but the usual way (`∇^2 f` is roughly `∇f ∇f^T`) will always give positive definite matrices.

PR #268, opened 2025-11-14
switch to using ReTest

luke-kiernan:
Currently, whenever you try to run the tests interactively (e.g. via `include("test/runtests.jl")`), you'll get an error: "`AAFactorization` not defined in `AppleAccelerate`." It's a strange side effect of some upstream changes in PNM. Switch tests to using `ReTest.jl`, which resolves the problem.

### DIFF
--- a/src/PowerFlowData.jl
+++ b/src/PowerFlowData.jl
@@ -154,9 +154,7 @@ const vPTDFPowerFlowData = PowerFlowData{
     PNM.VirtualPTDF{
         Tuple{Vector{Tuple{Int, Int}}, Vector{Int64}},
         Tuple{Dict{Tuple{Int, Int}, Int64}, Dict{Int64, Int64}},
-        PNM.KLU.KLUFactorization{Float64, Int64}, # TODO: could be AAFactorization.
-        # but for some reason it errors if I switch to 
-        # <:Union{PNM.KLU.KLUFactorization{Float64, Int64}, PNM.AAFactorization{Float64}
+        PNM.KLU.KLUFactorization{Float64, Int64},
     },
     PNM.ABA_Matrix{
         Tuple{Vector{Int64}, Vector{Int64}},


luke-kiernan:
I've run the tests locally, and the only 2 that are failing are the usual WECC 240 bus ones. [I'll troubleshoot that test failure tomorrow, now that I've resolved the PNM-PF compatibility issues.]

luke-kiernan:
> Hmm, uh, I'm not necessarily opposed to considering ReTest, but this seems like a weird reason to do it. Right now we use it for IS tests and not for anything else, and right now I find the IS tests a lot harder to work with than all the others.

I didn't exactly want to re-write everything in ReTest, either. However, I wasn't able to find a simpler solution. I tried the usual stuff (clearing the cache, re-compiling, etc.) but to no avail.

Specific nitpick: with `ReTest.jl`, I can't write `@testset "Test Foo with $item" for item in FooValues`. I'd welcome a better workaround than my current approach, which is:
```julia
function foo_test_body(item)
    [tests that apply to all items in FooValues]
end

testset "Test Foo" begin # manually unroll FooValues
    foo_test_body(item1)
    foo_test_body(item2)
    ...
end
```

jd-lara:
@GabrielKS we are testing still the re-test solution in the smaller packages. In this case, the whole issue with AA is just inexplicable because it only happens in the tests when we run it with that format, it doesn't happen when using PNM or anywhere else.

daniel-thom:
> adding @daniel-thom because he will definitely have opinions on this

Yes, I found ReTest.jl to be the least bad option. I use it because it allows me to quickly run a single test over and over (or use a regex matching specific test). This is how I iterate while developing code. Instructions are [here](https://github.com/NREL-Sienna/InfrastructureSystems.jl/blob/main/docs/src/dev_guide/tests.md)

PR #270, opened 2025-11-18
error if not in system base; set units in roundtrip test

luke-kiernan:
Continuation of PR #260. Instead of inserting `with_units_base` in a dozen spots, I thought it'd be cleaner to make things error if the units aren't `SYSTEM_BASE`.


GabrielKS (commented):
What's the rationale for imposing this somewhat arbitrary interface requirement on the user? Ultimately I think the statefulness of our units management system should be abolished (hence my VAM project) but in the meantime we can at least make it as harmless as possible.

GabrielKS (approved):
LGTM given change of error to assert and clarification that users should never encounter this, assuming tests are still passing locally.

PR #273, opened 2025-12-02
add basic performance test

m-bossart:
Adds performance test for `PowerFlows.jl `Follows the general structure of performance tests in `PowerNetworkMatrices.jl `and `PowerSimulations.jl`
- This tests the solve time but not the robustness of solvers to the initial conditions. 

@luke-kiernan Let me know if there are other tests that you think should be included here.

### DIFF
--- a/test/performance/performance_test.jl
+++ b/test/performance/performance_test.jl
@@ -21,7 +21,6 @@ for (group, name) in systems
         try
             pf = ACPowerFlow{solver}()
             pf_data = PF.PowerFlowData(pf, sys; correct_bustypes = true)
-            n = size(PF.get_bus_type(pf_data), 1)
             _, time_solve_1, _, _ = @timed solve_powerflow!(pf_data; pf = pf)
             open("solve_time.txt", "a") do io
                 write(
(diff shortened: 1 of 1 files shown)


test/performance/performance_test.jl
```diff
@@ -0,0 +1,49 @@
...
+            _, time_solve_1, _, _ = @timed solve_powerflow!(pf_data; pf = pf)
+            open("solve_time.txt", "a") do io
+                write(
+                    io,
+                    "| $(ARGS[1])-$(name)-$(solver)- first solve | $(time_solve_1) |\n",
+                )
+            end
+            pf = ACPowerFlow{solver}()
+            pf_data = PF.PowerFlowData(pf, sys; correct_bustypes = true)
+            n = size(PF.get_bus_type(pf_data), 1)
```
luke-kiernan: `n` is defined, but I don't see it used anywhere.

luke-kiernan (changes requested):
Looks reasonable.

Experimentally, I ran
```julia
runtimes = Vector{Float64}()
sizehint!(runtimes, 10)
for i in 1:10
    pf = PF.ACPowerFlow()
    pf_data = PF.PowerFlowData(pf, sys; correct_bustypes = true)
    _, time_solve, _, _ = @timed solve_powerflow!(pf_data; pf = pf)
    push!(runtimes, time_solve)
end
```
and it seems pretty consistent: 1.3 for the first run, 0.05-0.06 for the rest.

luke-kiernan (approved):

PR #274, opened 2025-12-03
Lk/psy5 docs rebase

luke-kiernan:
Bring docstrings (mostly) up to date. A rebase of some prior work.

Things that merit discussion:
- The math-y docstrings now render nicely in markdown, but that means they aren't very readable as plaintext.
- `PowerFlowData` and `solve_powerflow!(data)` are exported, when they aren't really intended for public usage.
  - I could change this, but it's not straightforward with how we currently have functions grouped into files.


luke-kiernan:
Further changes:
- The "system update" version of `solve_powerflow!` is renamed to `solve_and_store_power_flow!`
- No longer exporting `PowerFlowData`, its type aliases, `solve_powerflow!`, and `write_results`.
- Separate reference out into a couple sub sections.
   - Some of these I included just to fix `make.jl`'s "n docstrings not included in the manual" errors. There's probably a way to tell documenter "this docstring is for dev use only, please ignore it," though...

src/psse_export.jl
```diff
@@ -57,7 +56,7 @@ const WINDING_GROUP_NUMBER_TO_DEGREES = Dict(
     11 => 30,             # GROUP_11: 30 Degrees
 )
 
-# Each of the groups in the PSS/E v33 standard
+# Each of the groups in the PSS/3 v33 standard
```
jd-lara: interesting change here...

Issue #244, opened 2025-09-17
Bus type checks and HVDC


luke-kiernan:
PSY treated terminal nodes of HVDC lines as PV buses. With moving those bus type checks from PSY to here in PF, we should probably do the same when we add support for HVDC lines.

PR #282, opened 2025-12-26
include HVDC endpoints in "must be PV"

luke-kiernan:
Close issue #244.

I put all endpoints of all types of HVDC lines as "must be PV," but if we want something more lenient--e.g. perhaps the from bus of an LCC being PQ is okay--then let me know and I'll change it.


luke-kiernan:
> There is no requirement for then end points of HVDC to be PV, this depends on the control approach. Where is the rationale for this coming from?

The linked issue explains the reasoning:

> PSY treated terminal nodes of HVDC lines as PV buses. With moving those bus type checks from PSY to here in PF, we should probably do the same when we add support for HVDC lines.

If that was in error, then go ahead and close this PR.

Issue #216, opened 2025-08-04
Implement performant multiperiod DC PowerFlow


jd-lara:
The multi period DC PowerFlow and branch flow checks need to be parallelized to improve performance during the power flow in the loop. Needs https://github.com/NREL-Sienna/PowerNetworkMatrices.jl/issues/168 to work correctly.

luke-kiernan:
> The multi period DC PowerFlow and branch flow checks need to be parallelized to improve performance during the power flow in the loop. Needs https://github.com/NREL-Sienna/PowerNetworkMatrices.jl/issues/168 to work correctly.

I don't follow. Two main questions:

First, how many time steps are we considering? If you do something like
```julia
data = PowerFlowData(DCPowerFlow(), sys; time_steps = N)
# data.bus_activepower_{injection/withdrawals} now have N columns: write 
# injection/withdrawal at ith timestep to ith column
solve_powerflow!(data)
```
then in `solve_powerflow!`, we're doing a single matrix multiplication and computing the angles for all N time steps at once. That's already parallel and performant, for ~100 timesteps. Now, for ~10,000 time steps, there might be some other things to try.

Second, what do you mean by "branch flow checks?" Is that check something that should be implemented here, or in PSI?

luke-kiernan:
Implemented 2 in PNM [230](https://github.com/NREL-Sienna/PowerNetworkMatrices.jl/pull/230) and PF #283. I'll move this to "in review" for now: 3 doesn't seem like super high priority. (The KLU factorization step in inherently non-parallel, so we'd need to move the data over to the GPU, do the GEMM, then move it back, which will incur a penalty.)

PR #283, opened 2025-12-30
Set BLAS backend upon loading

luke-kiernan:
See also PNM PR [230](https://github.com/NREL-Sienna/PowerNetworkMatrices.jl/pull/230). Slight oddity: PNM's `__init__()` happens first, so if that PR is also merged, users will see PNM's "you might want to use [backend]" message followed by PF's "using [backend]" message.

edit: I did a profiling comparison on locally, and the AA BLAS is faster than OpenBLAS [with the default 8 threads].

edit 2: this is a step towards addressing issue #216

### DIFF
--- a/src/PowerFlows.jl
+++ b/src/PowerFlows.jl
@@ -46,6 +46,26 @@ const IS = InfrastructureSystems
 const PSY = PowerSystems
 const PNM = PowerNetworkMatrices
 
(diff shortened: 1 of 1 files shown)


jd-lara (changes requested):
Closing to handle this with preferences

Issue #265, opened 2025-11-11
Define PowerFlows and PowerSimulations interface


kennedy-mindermann:
Define the interface to support the integration of PowerFlows with PowerSimulations

luke-kiernan:
## Public
1. Solve and store to system [AC only]: `solve_and_store_power_flow!(sys, pf)`.
2. Solve and return `DataFrames`: `solve_power_flow(sys, pf)`. Returns a dictionary like `{"time1" => results1, "time2" => results2, ...}` where `results1` is a dictionary `{"bus_results" => DataFrame(...), "flow_results" => DataFrame(...)}`. For AC, `results1` might have a 3rd entry storing the LCC flows.

## Protected (PSI)
1. In-place on `PowerFlowData`: `solve_power_flow!(data, pf)`, maintaining single-argument ones for reverse-compatibility. [`solve_power_flow!(data::ACPowerFlowData)` defaults to `solve_power_flow(data, ACPowerFlow())`; similarly for DC varieties, just default-construct the matching `AbstractDCPowerFlow` subtype.]
2. `PowerFlowData` constructor: `PowerFlowData(sys, pf)`.


## Logic behind these choices

- name change for solve-and-store: don't want to make the in-place solve on `PowerFlowData` public.
- argument order change: mutated argument should be first for in-place, so might as well make it consistent, that the `System` or `PowerFlowData` is always the first arg.
- `pf` as positional arg (not kwarg): greater type stability for AC.
- `power_flow` over `powerflow`: consistency with `PowerFlows.jl`

Last 3 bullet points are somewhat subjective...but imo we should make these things consistent while we can.

There is also a DC powerflow `solve_powerflow(data, sys)` which returns data frames , but  we should probably just remove it. It isn't used anywhere.

luke-kiernan:
I'm realizing there's possibly more to do here. Here's the current state of things:

<img width="659" height="622" alt="Image" src="https://github.com/user-attachments/assets/05fda828-65eb-4cdd-8253-0d7dfb7a09e3" />

Synchronous condensers used to be neglected on the left path. Rodrigo is working on a PR to remedy that. However, I'm still a bit uneasy. Main concerns, in order of importance:

1. Are there other components handled differently on the left path vs. the right path? For each component type, is the same number added to the same field, regardless of whether it comes from the component's field or the optimization problem's solution?
2. Is `clear_injection_data!` zeroing out the right fields, no more and no less?
[...]

PR #289, opened 2026-01-09
add more fields to PFEM; add PFEM as field of PFD

luke-kiernan:
Resolve issue #265 and PSI [#284](https://github.com/NREL-Sienna/PowerFlows.jl/pull/284). Changes:

Add the following fields to the DC and AC solver types. The first 4 are for the `PowerFlowData` constructor and the last for `solve_powerflow!`.
- `network_reductions`
- `time_steps`
- `timestep_names`
- `correct_bustypes`
- `solver_kwargs`

Add the solver as a type parameter and a field of `PowerFlowData`. This will fix the above linked PSI issue. And we need most of the solver's fields anyway, so it's convenient to have it as a field.

I also removed the duplicate fields from `PowerFlowData`: **this last part does break backwards compatibility**. `PowerFlowData(ACPowerFlow(), sys; correct_bustypes = true)` now errors; instead, you need `PowerFlowData(ACPowerFlow(correct_bustypes = true), sys)`.  But we're doing UI changes here, so that's inevitable to some degree.

I plan to make the naming more uniform and switch to a single argument `solve_powerflow!`, but I didn't want to do too much in one PR. 

edit: I renamed the field `generator_slack_participation_factors` inside `PowerFlowData` to `computed_gspf` to avoid confusion with the `pf.generator_slack_participation_factors` field. The first is user input (could be `nothing`) whereas the second is some weights that we compute from the user input (always of type `Vector{Dict{Tuple{DataType, String}, Float64}}`).

### DIFF
--- a/src/ac_power_flow_jacobian.jl
+++ b/src/ac_power_flow_jacobian.jl
@@ -680,17 +680,10 @@ function _calculate_loss_factors(
     ref_mask = data.bus_type[:, time_step] .== (PSY.ACBusTypes.REF,)
     if count(ref_mask) > 1
         error(
-            "Loss factors with multiple REF buses isn't supported."
+            "Loss factors with multiple REF buses isn't supported.",
         )
     end
     pvpq_mask = .!ref_mask
-    ref = bus_numbers[ref_mask]
-    pvpq = bus_numbers[pvpq_mask]
-    pvpq_coords = J_INDEX_TYPE[]
-    for i in pvpq
-        push!(pvpq_coords, 2 * i - 1)  # 2x - 1
-        push!(pvpq_coords, 2 * i)      # 2x
-    end
     ref = findfirst(ref_mask)
     new_ref_mask = falses(size(ref_mask))
     new_ref_mask[ref] = true
--- a/test/test_dc_powerflow.jl
+++ b/test/test_dc_powerflow.jl
@@ -134,7 +134,7 @@ end
 @testset "DC power flow: StandardLoad" begin
     sys = PSB.build_system(PSB.PSITestSystems, "c_sys5")
     # change all loads to StandardLoad
-    dc_baseline = solve_powerflow(DCPowerFlow(), sys; correct_bustypes = true)
+    dc_baseline = solve_powerflow(DCPowerFlow(; correct_bustypes = true), sys)
     set_units_base_system!(sys, PSY.UnitSystem.NATURAL_UNITS)
     load = first(get_components(PowerLoad, sys))
     P = PSY.get_active_power(load)
--- a/test/test_loss_factors.jl
+++ b/test/test_loss_factors.jl
@@ -84,7 +84,7 @@ end
             _add_simple_line!(sys, b5, b6, 0.015, 0.08, 0.01)
             _add_simple_line!(sys, b4, b6, 0.012, 0.06, 0.015)
 
-            pf_lf = ACPowerFlow(ACSolver; calculate_loss_factors = true)
+            pf_lf = ACPowerFlow{ACSolver}(; calculate_loss_factors = true)
             data_loss_factors = PowerFlowData(pf_lf, sys)
 
             # Verify we have multiple REF buses before solving
@@ -92,13 +92,7 @@ end
             @test length(ref_buses) == 2
 
             # Solving should succeed (with a warning about multiple REF buses)
-            solve_powerflow!(data_loss_factors; pf = pf_lf)
-
-            # First REF bus should have loss factor of 1.0
-            @test data_loss_factors.loss_factors[first(ref_buses), 1] ≈ 1.0
-
-            # Loss factors should be computed (not NaN) for all buses
-            @test all(.!isnan.(data_loss_factors.loss_factors[:, 1]))
+            @test_throws ErrorException solve_powerflow!(data_loss_factors; pf = pf_lf)
         end
     end
 end
--- a/test/test_psse_export.jl
+++ b/test/test_psse_export.jl
@@ -310,8 +310,6 @@ function test_psse_round_trip(
 )
     raw_path, metadata_path =
         get_psse_export_paths(joinpath(export_location, scenario_name))
-    @test !isfile(raw_path)
-    @test !isfile(metadata_path)
 
     write_export(exporter, scenario_name; overwrite = true)
     @test isfile(raw_path)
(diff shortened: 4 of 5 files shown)


src/PowerFlowData.jl
```diff
@@ -92,9 +92,11 @@ the respective type of power flow evaluations.
 - `neighbors::Vector{Set{Int}}`: Vector with the sets of adjacent buses.
 """
 struct PowerFlowData{
+    T <: PowerFlowEvaluationModel,
     M <: PNM.PowerNetworkMatrix,
     N <: Union{PNM.PowerNetworkMatrix, Nothing},
 } <: PowerFlowContainer
+    pf::T
```
GabrielKS:
```suggestion
    pfem::T
```
feel like there's enough name confusion in this package as it is, we should be explicit when that's straightforward

src/PowerFlowData.jl
```diff
@@ -415,70 +418,52 @@ function make_and_initialize_powerflow_data(
...
 WARNING: functions for the evaluation of the multi-period AC PF still to be implemented.
 """
 function PowerFlowData(
     pf::ACPowerFlow{<:ACPowerFlowSolverType},
-    sys::PSY.System;
-    network_reductions::Vector{PNM.NetworkReduction} = Vector{PNM.NetworkReduction}(),
-    time_steps::Int = 1,
-    timestep_names::Vector{String} = String[],
-    correct_bustypes::Bool = false,
+    sys::PSY.System,
```
GabrielKS: nitpick, feel free to defer to a later PR: if you're sticking with [your proposal](https://github.com/NREL-Sienna/PowerFlows.jl/issues/265) to put the system/data first and the PFEM second in the public interface, would be nice to do that here too for consistency

src/powerflow_types.jl
```diff
@@ -58,41 +87,15 @@ struct ACPowerFlow{ACSolver <: ACPowerFlowSolverType} <: PowerFlowEvaluationMode
...
-- `calculate_loss_factors::Bool`: Whether to calculate loss factors during the power flow solution.
-    Default is `false`.
-- `generator_slack_participation_factors::Union{Nothing, Dict{Tuple{DataType, String}, Float64}, Vector{Dict{Tuple{DataType, String}, Float64}}}`:
-    An optional parameter that specifies the participation factors for generator slack in the power flow solution.
-    If `nothing`, all slack is picked up by the reference bus. If a `Dict`, it should map `(component_type, component_name)`
-    tuples to participation factors. If a `Vector`, it should contain multiple such dictionaries, 
-    allowing for different participation factors for different time steps.
-"""
-ACPowerFlow{ACSolver}(;
+# Keyword constructor for parametric type
```
GabrielKS: does the type parameter keep you from using `@kwdef`?

src/powerflow_types.jl
```diff
@@ -115,45 +123,37 @@ ACPowerFlow{ACSolver}(;
...
-        Nothing,
-        Dict{Tuple{DataType, String}, Float64},
-        Vector{Dict{Tuple{DataType, String}, Float64}},
-    } = nothing,
-    enhanced_flat_start::Bool = true,
-    robust_power_flow::Bool = false,
-    skip_redistribution::Bool = false,
-)
-    if calculate_loss_factors && ACSolver == LevenbergMarquardtACPowerFlow
+# Convenience constructor that allows specifying solver type as positional argument
```
GabrielKS: If this is just for convenience, why is it the only place that the LM validation happens?

src/solve_ac_powerflow.jl
```diff
@@ -5,14 +5,16 @@ Solves the power flow in the system and writes the solution into the relevant st
...
 
-Supports passing kwargs to the PF solver.
+Configuration options like `time_steps`, `time_step_names`, `network_reductions`, and
+`correct_bustypes` should be set on the `ACPowerFlow` object.
 
 The bus types can be changed from PV to PQ if the reactive power limits are violated.
 
 # Arguments
-- [`pf::ACPowerFlow{<:ACPowerFlowSolverType}`](@ref ACPowerFlow): The power flow solver instance.
+- [`pf::ACPowerFlow{<:ACPowerFlowSolverType}`](@ref ACPowerFlow): The power flow solver instance,
```
GabrielKS: I don't think it's correct to call this a solver instance

test/test_reduced_dc_powerflow.jl
```diff
@@ -14,19 +14,15 @@ dc_reduction_types = Dict{String, Vector{PNM.NetworkReduction}}(
...
-                dc_pf,
-                sys;
-                network_reductions = deepcopy(v),
-                correct_bustypes = true,
-            )
+            dc_pf_reduced =
+                pf_type(; network_reductions = deepcopy(v), correct_bustypes = true)
+            validate_reduced_powerflow(dc_pf_reduced, sys, v, unreduced)
+            results = PF.solve_powerflow(dc_pf_reduced, sys)
             # no write-results-to-system solve_and_store_power_flow! for DC: should we add one?
```
GabrielKS: (like, probably…)

src/powerflow_types.jl
luke-kiernan: Yeah. I tried to re-write using `@kwdef`, but two function definitions collide, leading to precompile errors.

By asking AI to generate a minimal example with similar behavior, I pinpointed the precise issue: `@kwdef` generates `ACPowerFlow(; kwargs...)`, which collides with our default-type-parameter-choice `ACPowerFlow(; kwargs...) = ACPowerFlow{Newton}(; kwargs...)`

luke-kiernan:
@GabrielKS I'm deferring cleaning up the name choices and the call signatures to a later PR, so I'll resolve all your comments related to that. Otherwise the git diff ends up being massive and it's hard to differentiate between the functional ones and the UI/syntax-only ones.

src/powerflow_types.jl
luke-kiernan: Good point.

Also, the design choice seems kind of silly: is it really that much more covenient to write `ACPowerFlow(FooSolver; ...)` than `ACPowerFlow{FooSolver}(; ...)`? I'll change it back

src/solve_ac_powerflow.jl
luke-kiernan: Sure. It's an instance of `ACPowerFlow`, not `ACPowerFlowSolverType`.

jd-lara:
@luke-kiernan resolve conflicts to finish review

PR #296, opened 2026-01-27
Lk/issue 265

luke-kiernan:
Rebased version of PR #289: resolves issue #265.

Move a bunch of fields that we currently pass to the `PowerFlowData` constructor into the solver, so that they're available for the user. But then if we're going to do that, we might as well make the solver a field of `PowerFlowData`, instead of storing everything in two places.

### DIFF
--- a/src/PowerFlowData.jl
+++ b/src/PowerFlowData.jl
@@ -107,7 +107,9 @@ struct PowerFlowData{
     bus_reactive_power_constant_impedance_withdrawals::Matrix{Float64}
     bus_reactive_power_bounds::Matrix{Tuple{Float64, Float64}}
     bus_slack_participation_factors::SparseMatrixCSC{Float64, Int}
-    computed_gspf::Vector{Dict{Tuple{DataType, String}, Float64}}
+    computed_generator_slack_participation_factors::Vector{
+        Dict{Tuple{DataType, String}, Float64},
+    }
     bus_type::Matrix{PSY.ACBusTypes}
     bus_magnitude::Matrix{Float64}
     bus_angles::Matrix{Float64}
@@ -206,7 +208,7 @@ get_voltage_stability_factors(pfd::PowerFlowData) = pfd.voltage_stability_factor
 
 # Field getter for expanded slack participation factors (one dict per time step)
 # Named "computed" to distinguish from the user-supplied pf.generator_slack_participation_factors
-get_computed_gspf(pfd::PowerFlowData) = pfd.computed_gspf
+get_computed_gspf(pfd::PowerFlowData) = pfd.computed_generator_slack_participation_factors
 
 # Delegating getters: delegate to the stored PowerFlowEvaluationModel
 get_calculate_loss_factors(pfd::PowerFlowData) = get_calculate_loss_factors(pfd.pf)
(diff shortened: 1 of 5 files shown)


src/power_flow_types.jl
```diff
@@ -56,13 +56,43 @@ See also: [`ACPowerFlow`](@ref).
...
+- `robust_power_flow::Bool`: Whether to use run a DC power flow as a fallback if the initial residual is large.
+    Default is `false`.
+- `skip_redistribution::Bool`: Whether to skip slack redistribution. Default is `false`.
+- `network_reductions::Vector{PNM.NetworkReduction}`: Network reductions to apply.
+    Default is an empty vector.
+- `time_steps::Int`: Number of time steps to solve. Default is `1`.
+- `time_step_names::Vector{String}`: Names for each time step. Default is an empty vector.
+- `correct_bustypes::Bool`: Whether to automatically correct bus types based on available generation.
+    Default is `false`.
+- `solver_kwargs::Dict{Symbol, Any}`: Additional keyword arguments to pass to the solver.
```
jd-lara: I don't love this use of a dict with symbol any for the solver parametrization and exposing in the public api the term kwargs.

src/power_flow_types.jl
```diff
@@ -76,6 +106,11 @@ struct ACPowerFlow{ACSolver <: ACPowerFlowSolverType} <: PowerFlowEvaluationMode
     enhanced_flat_start::Bool
     robust_power_flow::Bool
     skip_redistribution::Bool
+    network_reductions::Vector{PNM.NetworkReduction}
+    time_steps::Int
+    time_step_names::Vector{String}
+    correct_bustypes::Bool
+    solver_kwargs::Dict{Symbol, Any}
```
jd-lara: lets make this solver settings

src/PowerFlowData.jl
```diff
@@ -104,11 +106,8 @@ struct PowerFlowData{
     bus_active_power_constant_impedance_withdrawals::Matrix{Float64}
     bus_reactive_power_constant_impedance_withdrawals::Matrix{Float64}
     bus_reactive_power_bounds::Matrix{Tuple{Float64, Float64}}
-    generator_slack_participation_factors::Union{
-        Vector{Dict{Tuple{DataType, String}, Float64}},
-        Nothing,
-    }
     bus_slack_participation_factors::SparseMatrixCSC{Float64, Int}
+    computed_gspf::Vector{Dict{Tuple{DataType, String}, Float64}}
```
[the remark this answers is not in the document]
luke-kiernan: okay, I'll change it to `computed_generator_slack_participation_factors`

src/ac_power_flow_jacobian.jl
```diff
@@ -685,13 +685,6 @@ function _calculate_loss_factors(
         )
     end
     pvpq_mask = .!ref_mask
-    ref = bus_numbers[ref_mask]
```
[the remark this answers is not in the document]
luke-kiernan: It's leftover, from Roman's indexing based approach: the variables `pvpq_coords` and `pvpq` aren't actually used in what follows. `pvpq_coord_mask = repeat(pvpq_mask; inner = 2)` and `pvpq_mask` serve the same purpose.

src/power_flow_types.jl
jd-lara: I think these should be the solver_settings which is more specific to what it is.

PR #302, opened 2026-01-30
Fix tests for radial + ac convergence

m-bossart:
The fix in [PNM](https://github.com/JuliaTesting/ReTest.jl/issues/50) for modifying the Ybus after radial reduction appears to have fixed convergence issues for radial + AC in PowerFlows. This PR updates the tests to account for the cases that are now converging.


luke-kiernan (approved):
Looks good: ok to merge if CI succeeds.

PR #321, opened 2026-02-24
Tx Stepping Experiments

luke-kiernan:
wimp implementation of the Tx stepping method. Currently I have it converging on the ACTIVS 10k bus system in ~10 seconds. Yet to do performance and robustness experiments 

Currently not complatible with:
- HVDC lines
- network reductions
  - unclear to me if these are even compatible with the method. AI says: "The nice properties — trivial initial solution, guaranteed high-voltage convergence — depend on every physical line being individually shorted. The reduced equivalent doesn't faithfully represent that."
- PV-PQ switching and reactive power limits
- distributed slack
- loss factors


luke-kiernan:
experimental--closing

Issue #314, opened 2026-02-13
Post processing options for Branches for AC Power Flow


rodrigomha:
In some cases we need the line flows and line losses per arc, while in other cases per each component makes more sense.

Similar to https://github.com/NREL-Sienna/PowerFlows.jl/pull/306 for AC

PR #322, opened 2026-02-25
add arc vs branch option for post-processing

luke-kiernan:
Address issue #314

### DIFF
--- a/test/PowerFlowsTests.jl
+++ b/test/PowerFlowsTests.jl
@@ -44,7 +44,7 @@ const TEST_DATA_DIR = joinpath(
 )
 const DIFF_INF_TOLERANCE = 1e-4
 const DIFF_L2_TOLERANCE = 1e-3
-const TIGHT_TOLERANCE = 1e-9
+const TIGHT_TOLERANCE = 1e-7
 
 const LOG_FILE = "power-flows.log"
 


src/post_processing.jl
```diff
@@ -390,102 +436,157 @@ entries ``y_{tf}^{i+1}``, and superdiagonal entries ``y_{ft}^i``.
...
+    prev_bus_no = arc[1]
+    prev_V = V_endpoints[1]
+    for (i, segment) in enumerate(arc_entry)
+        (segment_from, segment_to) = PNM.get_arc_tuple(segment)
+        reversed = segment_from != prev_bus_no
+        current_bus_no = reversed ? segment_from : segment_to
+        current_V = (i == length(arc_entry)) ? V_endpoints[2] : x[i]
+
+        (V_from, V_to) = reversed ? (current_V, prev_V) : (prev_V, current_V)
+        if segment isa PNM.BranchesParallel
```
luke-kiernan: Code repetition: just call `_compute_segment_flows` on the segment instead.

src/post_processing.jl
```diff
@@ -390,102 +436,157 @@ entries ``y_{tf}^{i+1}``, and superdiagonal entries ``y_{ft}^i``.
...
+    _get_arc_endpoint_voltages(data, arc, time_step)
+
+Look up the complex voltages at the two endpoints of an arc from the solved power flow data.
+"""
+function _get_arc_endpoint_voltages(
+    data::ACPowerFlowData,
+    arc::Tuple{Int, Int},
+    time_step::Int,
+)
+    bus_lookup = get_bus_lookup(data)
```
luke-kiernan: Do I need to take the network reduction's bus map into account here?

test/test_post_processing.jl
```diff
@@ -19,33 +51,194 @@
...
-    # The datasets should be sorted by from bus, so order should also be the same (which is assumed in the test)
-    for (full_row, reduced_row) in
-        zip(eachrow(res_full_branch), eachrow(res_reduced_branch))
-        for name in names(full_row)
-            if name == "flow_name"
-                @test full_row[name] == reduced_row[name]
+    _compare_flow_dataframes(res_full_branch, res_reduced_branch)
+end
+
+@testset "AC power flow arc/branch flow reporting with degree-2 reduction" begin
```
m-bossart: The test included here are good, but I think we need another test that runs an AC Power Flow w/ and w/out degree-2 reductions and gets the same answer when reporting branch flows.

jd-lara:
@luke-kiernan and @m-bossart we should merge this along with the angle differences PR #326 so that we can report the angle differences also by arc or branch

test/test_post_processing.jl
m-bossart: I agree that this is testing the network reduction, but that test doesn't actually call the post-processing code which was added. It also skips Parallel within Series reductions.

luke-kiernan: Okay I've added such tests now

PR #349, opened 2026-04-07
simplify hessian matrix structure initialization

luke-kiernan:
Currently there's some graph stuff for figuring out the sparse structure of the hessian of the homotopy function. I did some profiling a while back and found that simply computing `J^T J` is faster. Both give the same results: we want (2-or-fewer)-hop neighbors. Also using `J^T J`'s structure generalizes better: e.g. proportional-to-headroom slack.

### DIFF
--- a/src/RobustHomotopy/homotopy_hessian.jl
+++ b/src/RobustHomotopy/homotopy_hessian.jl
@@ -81,12 +81,19 @@ function HomotopyHessian(data::ACPowerFlowData, time_step::Int)
         pfResidual.subnetworks,
         time_step,
     )
-    # Allocate Hv with the sparsity pattern of J' * J. Sparse `*` preserves the
-    # structural pattern regardless of nzval, so we just zero out the result.
-    # The per-call assertion in `A_plus_eq_BT_B!` guards against any future
-    # change in Julia that would drop structural zeros here.
+    # Allocate Hv with the maximal sparsity pattern of J' * J. Sparse `*`
+    # currently preserves structural zeros, but that isn't a documented
+    # SparseArrays contract, so we defensively fill nzval with ones to force
+    # the maximal pattern. We then restore J.Jv's original nzval — some
+    # entries (e.g. LCC angle-constraint diagonals of 1.0) are set at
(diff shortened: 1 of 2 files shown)


luke-kiernan:
Copilot was right — verified that ac_power_flow_jacobian.jl:331-332 sets the LCC angle-constraint diagonals to 1.0 at structure creation, and `_update_jacobian_matrix_values!` never rewrites them, so the old `nonzeros(J.Jv) .= 0.0` would corrupt them on any system with LCC. Fixed via save/restore of `J.Jv`'s nzval in d9b58e6a.

edit: turns out the RH method as written isn't compatible with LCCs anyway...but it's a one-time cost, so I'll leave the save-and-restore.

luke-kiernan:
Turns out these safeguards aren't strictly necessary: at the moment, `H = J' * J` keeps structural zeros. However, nothing in the `SparseArrays` docs dictates that behavior, so best not to rely on it.

luke-kiernan:
I added a "no LCCs" guard on the RH method path and a basic test that exercises said guard.

jd-lara:
> I added a "no LCCs" guard on the RH method path and a basic test that exercises said guard.

The problem is going to the that in real systems like WECC and the EI there will be LCCs in the data.

luke-kiernan:
> The problem is going to the that in real systems like WECC and the EI there will be LCCs in the data.

Sure, I can work on an LCC-compatible robust homotopy method. But I'd still vote for merging this: throwing a human-readable error is improvement over "dimension mismatch."

edit: there's a couple other things in this PR too, besides the LCC error

jd-lara:
> > The problem is going to the that in real systems like WECC and the EI there will be LCCs in the data.
> 
> Sure, I can work on an LCC-compatible robust homotopy method. But I'd still vote for merging this: throwing a human-readable error is improvement over "dimension mismatch."
> 
> edit: there's a couple other things in this PR too, besides the LCC error

Yeah, I don't think is blocking. Just nothing that it will be a consideration

jd-lara (approved):

PR #351, opened 2026-04-08
prep for contingencies

jd-lara:
I asked Claude to refactor the way to handle the internal PowerFlowData to better manage contingencies. 

This was the prompt 

```
The design idea is to extract the     
  results matrices defined in PowerFlowData into two different structs.                                                    
                                                                                                                           
  1. TimePowerFlowData which essentially retains the existing structure with matrices for all the results and              
  2. TimeContingencyPowerFlowData which now contains a tensor and a proper index to link the contingencies and time steps  of each value. The TimeContingencyPowerFlowData should also have placeholder to store the YbusModifications from each contingency as defined in ../PowerNetworkMatrices to facilitate the calculation of the updated Ybus used for each contingency.

  Make a plan to implement these new structs in a very performant way and make sure all the API for PowerFlowData without  contingencies works properly
  ```

This relies on code from PNM that isn't released yet

### DIFF
--- a/src/PowerFlowData.jl
+++ b/src/PowerFlowData.jl
@@ -59,22 +59,12 @@ the respective type of power flow evaluations.
 - `bus_reactive_power_bounds::Matrix{Float64}`:
         matrix containing upper and lower bounds for the reactive supply at each
         bus at each time period.
-- `bus_type::Matrix{PSY.ACBusTypes}`:
-        matrix containing type of buses present in the system.
-- `bus_magnitude::Matrix{Float64}`:
-        matrix containing the bus voltage magnitudes.
-- `bus_angles::Matrix{Float64}`:
-        matrix containing the bus voltage angles.
-- `arc_active_power_flow_from_to::Matrix{Float64}`:
-        matrix containing the active power flows measured at the `from` bus.
-- `arc_reactive_power_flow_from_to::Matrix{Float64}`:
-        matrix containing the reactive power flows measured at the `from` bus.
-- `arc_active_power_flow_to_from::Matrix{Float64}`:
-        matrix containing the active power flows measured at the `to` bus.
-- `arc_reactive_power_flow_to_from::Matrix{Float64}`:
-        matrix containing the reactive power flows measured at the `to` bus.
-- `arc_angle_differences::Matrix{Float64}`:
-        matrix containing the voltage angle difference (θ_from − θ_to) across each arc.
+- `results::R`:
+        result storage container (see [`TimePowerFlowData`](@ref) or
+        [`TimeContingencyPowerFlowData`](@ref)). Result fields such as `bus_magnitude`,
+        `bus_angles`, `bus_type`, arc flows, `converged`, `loss_factors`, etc. are
+        stored here but accessible directly via `data.bus_magnitude` through property
+        forwarding.
 - `generic_hvdc_flows::Dict{Tuple{Int, Int}, Tuple{Float64, Float64}}`:
         dictionary mapping each generic HVDC line (represented as a tuple of the from and to bus
         numbers) to a tuple of `(P_from_to, P_to_from)` active power flows.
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

File: src/power_flow_results.jl [hunk 1]
```diff
@@ -0,0 +1,249 @@
...
+- `arc_reactive_power_flow_from_to::Matrix{Float64}`: reactive power flow measured at the from bus.
+- `arc_active_power_flow_to_from::Matrix{Float64}`: active power flow measured at the to bus.
+- `arc_reactive_power_flow_to_from::Matrix{Float64}`: reactive power flow measured at the to bus.
+- `arc_angle_differences::Matrix{Float64}`: voltage angle difference across each arc.
+- `converged::BitVector`: convergence status for each time step.
+- `loss_factors::Union{Matrix{Float64}, Nothing}`: bus loss factors, or `nothing` if not computed.
+- `voltage_stability_factors::Union{Matrix{Float64}, Nothing}`: voltage stability factors, or `nothing` if not computed.
+- `arc_active_power_losses::Union{Matrix{Float64}, Nothing}`: active power losses per arc, or `nothing` if not computed.
+"""
+struct TimePowerFlowData <: AbstractPowerFlowResults
```
luke-kiernan, line 58: What about bus active and reactive powers? At the very least, I'd add `bus_reactive_power_injections`, for the `Q` values at the PV buses. I think we should probably include all 4, bus active/reactive power injections/withdrawals.

LCCs are another consideration.

src/power_flow_results.jl
```diff
@@ -0,0 +1,249 @@
...
+    bus_magnitude::Matrix{Float64}
+    bus_angles::Matrix{Float64}
+    bus_type::Matrix{PSY.ACBusTypes}
```
luke-kiernan, line 61: Is the omission of `bus_axis` and `arc_axis` purposeful? They're not results, but in order to anything useful with the results, you kind of need those. Same goes for lookups.

src/power_flow_results.jl
```diff
@@ -0,0 +1,249 @@
...
+- `arc_angle_differences::Array{Float64, 3}`: voltage angle difference across each arc.
+- `converged::BitMatrix`: convergence status with dimensions `(time_steps, contingencies)`.
+- `loss_factors::Union{Array{Float64, 3}, Nothing}`: bus loss factors, or `nothing` if not computed.
+- `voltage_stability_factors::Union{Array{Float64, 3}, Nothing}`: voltage stability factors, or `nothing` if not computed.
+- `arc_active_power_losses::Union{Array{Float64, 3}, Nothing}`: active power losses per arc, or `nothing` if not computed.
+- `contingency_labels::Vector{String}`: human-readable labels for each contingency.
+- `contingency_lookup::Dict{String, Int}`: mapping from contingency label to index.
+- `network_modifications::Vector{Union{Nothing, PNM.NetworkModification}}`: network modification for each contingency.
+"""
+struct TimeContingencyPowerFlowData <: AbstractPowerFlowResults
```
luke-kiernan, line 161: A bit repetitive: the first dozen fields here are the same as `TimePowerFlowData`, except with one additional dimension. Suggestions:
- Store a `Vector{TimePowerFlowData}`
- Make a supertype `PowerFlowResultArrays{N}` which stores `N`-dimensional arrays.

luke-kiernan (commented):
Overall, I like the direction this is going. A few thoughts on how you regrouped the fields

src/PowerFlowData.jl
```diff
@@ -116,28 +104,33 @@ struct PowerFlowData{
...
-    :converged,
-    :loss_factors,
-    :voltage_stability_factors,
-    :arc_active_power_losses,
-])
-
+# Uses an explicit === chain instead of `s in _RESULT_FIELD_NAMES` for type stability.
 function Base.getproperty(pfd::PowerFlowData, s::Symbol)
-    if s in _RESULT_FIELD_NAMES
+    if s === :bus_magnitude ||
```
luke-kiernan: Hmm. I'm mildly skeptical that this is type stable. Plus it's real ugly. Claude code suggests the following:
```julia
Base.@constprop :aggressive @inline function Base.getproperty(pfd::PowerFlowData, s::Symbol)
    results = getfield(pfd, :results)
    return hasfield(typeof(results), s) ?
        getfield(results, s) : 
        getfield(pfd, s)
end
```

PR #356, opened 2026-04-18
Add literate.jl docs infrastructure

kdayday:
- Adds Literate.jl tutorial processing infrastructure with make_tutorials.jl, including docs\ dependencies
- translates tutorial on solving a power flow into literate.jl format, and bug fixes existing errors on solving a DC power flow due to API changes 
    - besides format translation and basic bug fixes, does not include any of the other changes previously discussed with @luke-kiernan 
- Small docstring format fixes and reorganization of explanation section
- Add pre-commit yaml for formatter
- Add github action to remove docs previews


docs/make_tutorials.jl
```diff
@@ -0,0 +1,390 @@
...
+    "compat" => "Compat",
+    "todo" => "TODO",
+    "details" => "Details",
+)
+
+# Preprocess Literate source to convert Documenter-style admonitions into Jupyter-friendly
+# blockquotes. Used only for notebook output; markdown keeps `!!! type` and is rendered by
+# Documenter. Admonitions are not recognized by common mark or Jupyter; see
+# https://fredrikekre.github.io/Literate.jl/v2/tips/#admonitions-compatibility
+function preprocess_admonitions_for_notebook(str::AbstractString)
```
luke-kiernan, line 163: This function has enough going on that I'd consider writing some test cases.

docs/make_tutorials.jl
```diff
@@ -0,0 +1,390 @@
...
+    m = match(r"^(#+ .+)$"m, content)
+    if m !== nothing
+        heading = m.match
+        content = replace(content, r"^(#+ .+)$"m => heading * download_section; count = 1)
+    end
+    return content
+end
+
+# Function to add Pkg.status() to notebook within the first markdown cell
+function add_pkg_status_to_notebook(nb::Dict)
```
luke-kiernan, line 246: Same here too.

docs/make_tutorials.jl
```diff
@@ -0,0 +1,390 @@
...
+    # Call PrettyTables directly with row/column limits
+    # This ensures only 10 rows are shown regardless of DataFrame size
+    PrettyTables.pretty_table(io, df;
+        backend = :text,
+        maximum_number_of_rows = 10,
+        maximum_number_of_columns = 80,
+        show_omitted_cell_summary = true,
+        compact_printing = false,
+        limit_printing = true)
+end
```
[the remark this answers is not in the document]
luke-kiernan: Claude claims that you can just do `ENV["DATAFRAMES_ROWS"] = "10"` and `ENV["DATAFRAMES_COLUMNS"] = "80"` at the top of the build script and get pretty much the same effect, though you'd be missing the other options like `compact_printing`.

luke-kiernan (changes requested):
`docs/make_tutorials.jl` and all its regex manipulations looks difficult to understand and maintain.

One idea: switch to using [Commonmark.jl](https://michaelhatherly.github.io/CommonMark.jl/stable/). Walk the document as a tree, instead of as text. `add_download_links` is basically "after each image node, add a link", easily done via [stuff like this](https://michaelhatherly.github.io/CommonMark.jl/stable/ast/#Tree-Manipulation). They even have something for [admonitions](https://michaelhatherly.github.io/CommonMark.jl/stable/extensions/#Admonitions)! AI could probably do a lot of the heavy lifting.

The `Base.show` override for `DataFrame` here is a little iffy type piracy wise.

If you're going to keep it as-is, I'd write some tests and/or document the intended functionality better, e.g. sample input/output before each function.

rodrigomha:
> `docs/make_tutorials.jl` and all its regex manipulations looks difficult to understand and maintain.
> 
> One idea: switch to using [Commonmark.jl](https://michaelhatherly.github.io/CommonMark.jl/stable/). Walk the document as a tree, instead of as text. `add_download_links` is basically "after each image node, add a link", easily done via [stuff like this](https://michaelhatherly.github.io/CommonMark.jl/stable/ast/#Tree-Manipulation). They even have something for [admonitions](https://michaelhatherly.github.io/CommonMark.jl/stable/extensions/#Admonitions)! AI could probably do a lot of the heavy lifting.
> 
> The `Base.show` override for `DataFrame` here is a little iffy type piracy wise.
> 
> If you're going to keep it as-is, I'd write some tests and/or document the intended functionality better, e.g. sample input/output before each function.

We definitely want to stick with what we have right now. I think, running the documenter is the test itself. If any of these methods fail, the documenter build will fail, so although adding a test for this is a good idea, we don't really want to add more github actions for the documentation if we can.

docs/make_tutorials.jl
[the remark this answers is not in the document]
kdayday: I've tried the environment variable thing multiple times, and it doesn't actually work. More from AI: "Those variables are not part of the standard DataFrames display control path in a way that reliably affects docs builds. In docs generation, output usually goes through non-interactive IO contexts where row/column limits are controlled by:
- `show(...; allrows=false, allcols=false, ...)`
- IOContext flags / display size
- pretty-print backend options (e.g. `PrettyTables` kwargs)"

I implemented `kwargs` in the current `Base.show`

docs/make_tutorials.jl
kdayday: I did not add tests, but did add to the documentation at the top of each function

kdayday:
> > `docs/make_tutorials.jl` and all its regex manipulations looks difficult to understand and maintain.
> > One idea: switch to using [Commonmark.jl](https://michaelhatherly.github.io/CommonMark.jl/stable/). Walk the document as a tree, instead of as text. `add_download_links` is basically "after each image node, add a link", easily done via [stuff like this](https://michaelhatherly.github.io/CommonMark.jl/stable/ast/#Tree-Manipulation). They even have something for [admonitions](https://michaelhatherly.github.io/CommonMark.jl/stable/extensions/#Admonitions)! AI could probably do a lot of the heavy lifting.
> > The `Base.show` override for `DataFrame` here is a little iffy type piracy wise.
> > If you're going to keep it as-is, I'd write some tests and/or document the intended functionality better, e.g. sample input/output before each function.
> 
> We definitely want to stick with what we have right now. I think, running the documenter is the test itself. If any of these methods fail, the documenter build will fail, so although adding a test for this is a good idea, we don't really want to add more github actions for the documentation if we can.

@luke-kiernan @rodrigomha I didn't add tests, but I've updated the functional issues copilot identified and added some input/output documentation for the functions. I think the outstanding question is about the `Base.show` override. The issue is that the output of `@example` blocks can print the entirety of large tables, unlike the REPL. I'd like an automated way to truncate outputs so users can get a taste, but tutorial writers don't need to remember to call `show` or `first` (because I don't trust us to remember or fully look through all tutorials, even though that's ideal).  Setting environment variables doesn't work as an alternative. I've at least implemented the `kwargs` option that copilot suggested.

luke-kiernan (approved):
Much improved.

Didn't try running myself locally, but if it works for you and on CI, then I'm ok with merging.

PR #363, opened 2026-05-06
[DO NOT MERGE] Repro: Windows-only BoundsError in legacy PF loss factors

luke-kiernan:
Draft PR to reproduce a Windows-only `BoundsError` originally seen on the `windows-bug` branch of luke-kiernan/PowerFlows.jl (May 2025).

The original branch from a year ago had massive merge conflicts with current `main`, and main's `test/test_utils/legacy_pf.jl` already includes the `_legacy_dSbus_dV` / `_legacy_J` recompute that was the original Windows workaround. So instead, this branch is rebased onto current `main` with a single targeted commit that comments out those two recomputes inside the `get_calculate_loss_factors` block.

**Expected:** Windows job fails with `BoundsError`; macOS and Linux pass.

**Goal:** confirm the bug still reproduces on current `main` + current registered dependency stack before bisecting KLU.jl / SuiteSparse_jll.

Not for merge.

### DIFF
--- a/.github/workflows/pr_testing.yml
+++ b/.github/workflows/pr_testing.yml
@@ -10,7 +10,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        julia-version: ['1']
+        julia-version: ['1.11']
         julia-arch: [x64]
         os: [ubuntu-latest, windows-latest, macOS-latest]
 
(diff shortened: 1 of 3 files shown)


luke-kiernan:
Closing — bug no longer reproduces on current main + deps. Reverting Roman's PR #93 (spdiagm→Diagonal) alone does not bring the BoundsError back, so the silent fix likely came from KLU.jl / SuiteSparse_jll upgrades rather than from the in-tree refactor.

PR #364, opened 2026-05-12
Rectangular Power flow

jd-lara:
Requires the KLU branch from PNM. Open for the first round of feedback


src/lcc_utils.jl
```diff
@@ -70,33 +70,50 @@ function _calculate_dQ_dα_lcc(
...
 end
 
-function _update_ybus_lcc!(data::PowerFlowData, time_step::Int64)
+"""
+    _update_ybus_lcc!(data, time_step; vm_fn = (i, ts) -> data.bus_magnitude[i, ts])
+
+Recompute `data.lcc.rectifier.phi`, `data.lcc.inverter.phi`, and
+`data.lcc.branch_admittances` for each LCC at `time_step`. By default the
+voltage magnitude at each AC terminal is read from `data.bus_magnitude`,
+which is correct for the polar formulation. The `vm_fn(bus_ix, time_step) ->
```
luke-kiernan: If we only have 2 cases we're interested in, passing a full function as a keyword argument seems like overkill.

src/power_flow_method.jl
```diff
@@ -760,3 +760,122 @@ function _newton_power_flow(
...
+          "$(norm(residual.Rv, 2)) L2, " *
+          "$(norm(residual.Rv, Inf)) L∞"
+    J = ACRectangularCIJacobian(residual, time_step)
+    converged = norm(residual.Rv, Inf) < tol
+    i = 0
+    if !converged
+        linSolveCache = KLULinSolveCache(J.Jv)
+        symbolic_factor!(linSolveCache, J.Jv)
+        stateVector = StateVectorCache(x0_computed, residual.Rv)
+        T_strategy = if step_strategy == :trust_region
```
luke-kiernan: nitpick: using a keyword argument that's passed as a symbol here means type instability.

src/ac_power_flow_residual.jl
```diff
@@ -464,3 +450,49 @@ function _find_subnetworks_for_reference_buses(
...
+
+"""
+    _build_bus_slack_participation_factors(data, bus_type, subnetworks, time_step)
+
+Collect the per-bus generator-slack-participation factors (REF and PV buses
+only), validate that the sum is positive and no value is negative, and
+normalize so that each subnetwork's participating buses sum to 1. Returns
+a `SparseVector{Float64, Int}` of length `n_buses`.
+
+Shared between the polar `ACPowerFlowResidual` and rectangular CI residual
```
luke-kiernan: What is CI here?

src/rectangular_ci_power_flow_residual.jl
```diff
@@ -0,0 +1,326 @@
...
+        # At PV buses, data.bus_magnitude holds V_set (not the state magnitude).
+        # Override the magnitude provider so LCC math sees |V_state|, keeping the
+        # residual consistent with the Jacobian which operates on (e_state, f_state).
+        _update_ybus_lcc!(data, time_step;
+            vm_fn = (i, _) -> sqrt(e_state[i]^2 + f_state[i]^2))
+    end
+
+    # 2) Compute P_eff / Q_eff (slack distribution + ZIP constant-current correction).
+    @inbounds for i in 1:n_buses
+        # ZIP const-I uses |V_state| (state magnitude), not V_set.
```
luke-kiernan, line 206: What about constant impedance loads?

edit: handled in the Y bus, but a comment along those lines would be welcome

src/rectangular_ci_power_flow_residual.jl
```diff
@@ -0,0 +1,326 @@
...
+            # so the recovered injection matches polar's `bus_active_power_injections`
+            # (which includes `const_I * V_set` via `get_bus_active_power_total_withdrawals`).
+            Vm = sqrt(V_sq)
+            P_eff = P_gen - const_I_P[i] * Vm
+            Q_eff = Q_gen - const_I_Q[i] * Vm
+            F[off] += (P_eff * e_i + Q_eff * f_i) / V_sq
+            F[off + 1] += (P_eff * f_i - Q_eff * e_i) / V_sq
+        else
+            P_i = P_eff_cache[i]
+            Q_i = bt == PSY.ACBusTypes.PV ? Q_state[i] : Q_eff_cache[i]
```
luke-kiernan, line 270: Is this correct? What's stopping us from having a ZIP load at a PV bus? Or, if there is a ZIP load at a PV bus, why do we use `Q_state` over `Q_eff_cache`?

src/rectangular_ci_setup.jl
```diff
@@ -0,0 +1,168 @@
...
+    fold_zip_constant_z!(Y_bus_eff, data, time_step)
+
+Add the constant-impedance ZIP load components into the `Y_bus_eff` diagonal
+as fixed shunt admittances. Sienna's ZIP load model parameterizes the load at
+`|V| = 1.0 pu` directly: a load with `constant_impedance_active_power = β_P`
+and `constant_impedance_reactive_power = β_Q` draws `S = (β_P + jβ_Q)·|V|²`.
+As a shunt admittance this is `Y_sh = (β_P − jβ_Q)` because
+`|V|²·conj(Y_sh) = (β_P + jβ_Q)·|V|²`.
+"""
+function fold_zip_constant_z!(
```
luke-kiernan, line 41: Oh, here's where constant Z is handled. Modifying the Y bus is a little confusing, but I can accept it. If we're going to handle it this way for rectangular, I'd be in favor of doing the same for polar too.

src/rectangular_ci_power_flow_residual.jl
```diff
@@ -0,0 +1,326 @@
...
+    # Populate state caches before LCC admittance refresh (LCC needs |V_state|).
+    @inbounds for i in 1:n_buses
+        off = Int(bus_state_offset[i])
+        bt = bus_types[i]
+        if bt == PSY.ACBusTypes.REF
+            Vm = data.bus_magnitude[i, time_step]
+            θ = data.bus_angles[i, time_step]
+            e_state[i] = Vm * cos(θ)
+            f_state[i] = Vm * sin(θ)
+            Q_state[i] = NaN
```
luke-kiernan, line 189: Not a huge fan of using NaN as "shouldn't be read," but I guess it's fine.

src/rectangular_ci_power_flow_residual.jl
```diff
@@ -0,0 +1,326 @@
...
+        bt = bus_types[i]
+        e_i = e_state[i]
+        f_i = f_state[i]
+        V_sq = e_i^2 + f_i^2
+        if bt == PSY.ACBusTypes.REF
+            c_ref = bus_slack_participation_factors[i]
+            P_slack_total = x[off] - P_net_set[i]
+            P_gen = P_net_set[i] + c_ref * P_slack_total
+            Q_gen = x[off + 1]
+            # |V| at REF is fixed at V_set; subtract the ZIP constant-current draw
```
luke-kiernan, line 256: Yeah that's confusing. (Why are we including the withdrawals in the injections?) I'm okay with this for now, but I'll open an issue to clean this up.

src/rectangular_ci_power_flow_jacobian.jl
```diff
@@ -0,0 +1,606 @@
...
+            c_k == 0.0 && continue
+            bus_k == ref_bus && continue
+            k_off = Int(bus_state_offset[bus_k])
+            bt_k = bus_types[bus_k]
+            bt_k == PSY.ACBusTypes.REF && continue
+            e_k = e_state[bus_k]
+            f_k = f_state[bus_k]
+            V_sq = e_k^2 + f_k^2
+            Jv[k_off, ref_off] = c_k * e_k / V_sq
+            Jv[k_off + 1, ref_off] = c_k * f_k / V_sq
```
luke-kiernan: You're intersecting `subnetwork_buses` with the nonzeros of `bus_slack_participation_factors`, which is doable in a single linear pass if `subnetwork_buses` is sorted.

Also worth noting here that the inner loop here is all in one column, `Jv[:, ref_off]`. The compiler might be doing this already, but you could cache the row offsets for that column outside the loop. Further, you could do a linear pass on all 3 together...but I'll need to think more if that actually makes sense. (Can the column of `J` have significantly more nonzeros than the intersection of `subnetwork_buses` and the nonzeros in `bus_slack_participation_factors`?)

src/rectangular_ci_power_flow_jacobian.jl
```diff
@@ -0,0 +1,591 @@
...
+        end
+    end
```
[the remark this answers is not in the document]
luke-kiernan: Agreed, would be good to find a way to avoid the nonlinear access pattern of `Jv[off, off]`

src/rectangular_ci_power_flow_jacobian.jl
```diff
@@ -0,0 +1,591 @@
...
+                for c in 0:(n_cols_to_write - 1)
+                    push!(rows, J_INDEX_TYPE(row_off + r))
+                    push!(cols, J_INDEX_TYPE(col_off + c))
+                    push!(vals, 0.0)
+                end
+            end
+            # For PV columns (when row != col), we still need the Q column entries
+            # for the diagonal block (∂I_spec/∂Q at the PV bus itself). Those are
+            # captured by row == col case above. No additional entries needed off-diag.
+        end
```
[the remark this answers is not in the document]
luke-kiernan: Wouldn't you also get extra nonzero entries in the ref bus column from the state variables being different? Meaning that CoPilot here is incorrect. (Are we also updating those entries correctly?)

luke-kiernan (commented):
Some comments. I have yet to profile it locally: mostly big-picture stuff about how this algorithm differs.

I suppose we'll also need to disable or adapt the voltage stability factors for the rectangular formulation: our jacobian entries are no longer derivatives of power, but derivatives of current.

src/rectangular_ci_power_flow_residual.jl
jd-lara: see #365

src/rectangular_ci_setup.jl
jd-lara: see #365

src/rectangular_ci_power_flow_jacobian.jl
```diff
@@ -0,0 +1,591 @@
...
+            (col_e_tb, idx_alpha_i, 0.0),
+            (col_f_tb, idx_tap_i, 0.0),
+            (col_f_tb, idx_alpha_i, 0.0),
+            # LCC tail rows (F_t_r, F_t_i) × bus cols.
+            (idx_tap_r, col_e_fb, 0.0),
+            (idx_tap_r, col_f_fb, 0.0),
+            (idx_tap_i, col_e_fb, 0.0),
+            (idx_tap_i, col_f_fb, 0.0),
+            (idx_tap_i, col_e_tb, 0.0),
+            (idx_tap_i, col_f_tb, 0.0),
```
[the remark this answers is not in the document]
jd-lara: This is not totally correct, also we need a more consistent effort to verify the LCC behavior with different control flags

src/rectangular_ci_setup.jl
```diff
@@ -0,0 +1,168 @@
...
+            data.bus_magnitude[i, time_step] = sqrt(e^2 + f^2)
+            data.bus_angles[i, time_step] = atan(f, e)
+        else  # PV: do NOT overwrite bus_magnitude (V_set must be preserved).
+            e = x[off]
+            f = x[off + 1]
+            data.bus_angles[i, time_step] = atan(f, e)
+            Q_gen = x[off + 2]
+            data.bus_reactive_power_injections[i, time_step] =
+                Q_gen + data.bus_reactive_power_withdrawals[i, time_step]
+        end
```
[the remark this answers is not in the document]
jd-lara: same as above

src/rectangular_ci_power_flow_residual.jl
jd-lara: This is because at a PV bus the reactive variable  is the bus's net Q injection, treated as an unknown. The solver iterates until the PV voltage-magnitude row is satisfied, so at convergence Q_state already equals Q_gen. Since the voltage is fixed,  the ZIP current withdrawal is implicit in the value the solver converges to There's no separate quantity to add in those cases, this is a feature of being a current injection formulation

luke-kiernan (approved):

PR #366, opened 2026-05-15
Fix LCC Jacobian: correct dQ/dV, dQ/dt; add missing inverter-side entries

luke-kiernan:
## Summary
- **Fixed algebra bugs** in `_calculate_dQ_dV_lcc` and `_calculate_dQ_dt_lcc` (`src/lcc_utils.jl`): the prior formulas had a spurious extra `V` factor, `sin²(φ)` instead of `V·sin(φ)` in the denominator, and a missing `sign(I_dc)` factor. The derivation was inconsistent with `Q = V·t·(√6/π)·I_dc·sin(φ)` chained through the simplified `φ = arccos(sign(I_dc)·(cos(α) - x·I_dc/(√2·t·V)))`.
- **Added the missing inverter-side entries** to the LCC Jacobian (`src/ac_power_flow_jacobian.jl`): the rectifier-terminal block had all six bus-side derivatives (∂P/∂V, ∂P/∂t, ∂P/∂α, ∂Q/∂V, ∂Q/∂t, ∂Q/∂α), but the inverter mirror was entirely absent — both in the sparsity-structure builder and in the update path. Added all six, plus the four new cross entries (bus rows ↔ inverter-side state columns) to `_create_jacobian_matrix_structure_lcc`.
- **Guarded against the φ-clamp singularity**: when `sin(φ) → 0` (the `arccos` argument hits the `clamp(..., -1, 1)` boundary), the analytic derivative formulas diverge as 1/sin(φ), but the *true* derivative in the clamped regime is 0 (φ is locally pinned, residual locally constant). The Q-derivative helpers now short-circuit to 0 below `sin(φ) < 1e-8`. Without this, Newton-Raphson diverges on LCC systems whose initial state lies on the clamp.
- **Logged a warning** in `_calculate_ϕ_lcc` when the clamp activates, with `maxlog=5` to avoid spam in the hot path.
- **Re-enabled** the previously-commented `verify_jacobian` test on the LCC system (`test/test_jacobian.jl`): bumped the inverter extinction angle off the clamp boundary at the test point, switched to central differencing for better FD accuracy, and confirmed all 10 columns match analytically.
- **Updated** `docs/src/explanation/lcc_model.md` with the corrected ∂Q_r/∂V_r and ∂Q_r/∂t_r formulas and added the new inverter-side P/Q derivative entries.

## Test plan
- [x] `Jacobian verification with LCC` testset passes (all 10 state-vector columns agree with central-FD to rtol=1e-6).
- [x] `Test LCC consistency` and `Test LCC` testsets still pass (NR converges on the simple LCC system).
- [x] Full test suite: 41,449 passed, 0 failed, 0 errored.

## Notes on remaining model limitations
The simplified-φ definition (closed-form `arccos(...)`) is fundamentally non-smooth at the clamp boundary. Newton may still struggle on systems whose converged operating point lies right at the clamp (unusual but possible). A principled fix is the full overlap-angle commutation model (Panosyan thesis, cited in `lcc_model.md`); that's a larger modeling effort and out of scope here. The warning logged by `_calculate_ϕ_lcc` surfaces clamp activation when it happens.

### DIFF
--- a/src/ac_power_flow_jacobian.jl
+++ b/src/ac_power_flow_jacobian.jl
@@ -576,8 +576,23 @@ function _set_entries_for_lcc(data::ACPowerFlowData,
 
         s = _lcc_jacobian_scalars(data, i, time_step, Vm_fb, Vm_tb)
 
+        # True-φ ∂P/∂V and ∂P/∂t for the rectifier side. The polar
+        # α-approximation drops the `∂φ/∂Vm` (resp. `∂φ/∂t`) chain term —
+        # which vanishes only when `x_t = 0`. The true-φ formulas are
+        # singularity-free (the `sin(φ)` from `∂φ/∂(V,t)` cancels the
+        # `-sin(φ)` from differentiating `cos(φ)`).
+        # For ∂P/∂α the analogous cancellation makes the α-approximation
+        # exact, so those entries stay on `common_alpha_*`.
+        dP_dV_fb = _calculate_dP_dV_lcc(s.tap_r, s.i_dc, xtr_r, Vm_fb, phi_r)
+        dP_dt_fb = _calculate_dP_dt_lcc(s.tap_r, s.i_dc, xtr_r, Vm_fb, phi_r)
+        # Inverter side: `P_lcc_to = V_tb · tap_i · √6/π · I_dc · cos(phi_i)`
+        # with `phi_i` already encoding the inverter sign convention via
+        # `_calculate_ϕ_lcc(-I_dc, …)`. Pass the same positive `I_dc`.
+        dP_dV_tb = _calculate_dP_dV_lcc(s.tap_i, s.i_dc, xtr_i, Vm_tb, phi_i)
+        dP_dt_tb = _calculate_dP_dt_lcc(s.tap_i, s.i_dc, xtr_i, Vm_tb, phi_i)
+
         if bus_type_fb == PSY.ACBusTypes.PQ
-            Jv[idx_p_fb, idx_p_fb] += s.common_tap_r # ∂P_fb/∂V_fb
+            Jv[idx_p_fb, idx_p_fb] += dP_dV_fb # ∂P_fb/∂V_fb
             Jv[idx_q_fb, idx_p_fb] +=
                 _calculate_dQ_dV_lcc(s.tap_r, s.i_dc, xtr_r, Vm_fb, phi_r) # ∂Q_fb/∂V_fb
 
@@ -586,17 +601,17 @@ function _set_entries_for_lcc(data::ACPowerFlowData,
             Jv[idx_q_fb, idx_angle_from] =
                 _calculate_dQ_dα_lcc(s.tap_r, s.i_dc, xtr_r, Vm_fb, phi_r, alpha_r) # ∂Q_fb/∂α_fb
 
-            Jv[idx_tap_from, idx_p_fb] = s.common_tap_r # ∂F_t_fb/∂V_fb
-            Jv[idx_tap_to, idx_p_fb] = s.common_tap_r # ∂F_t_tb/∂V_fb
+            Jv[idx_tap_from, idx_p_fb] = dP_dV_fb # ∂F_t_fb/∂V_fb
+            Jv[idx_tap_to, idx_p_fb] = dP_dV_fb # ∂F_t_tb/∂V_fb
         end
 
         if bus_type_fb == PSY.ACBusTypes.PQ || bus_type_fb == PSY.ACBusTypes.PV
-            Jv[idx_p_fb, idx_tap_from] = s.common_fb * s.cos_alpha_r # ∂P_fb/∂t_fb
-            Jv[idx_p_fb, idx_angle_from] = s.common_alpha_r # ∂P_fb/∂α_fb
+            Jv[idx_p_fb, idx_tap_from] = dP_dt_fb # ∂P_fb/∂t_fb
+            Jv[idx_p_fb, idx_angle_from] = s.common_alpha_r # ∂P_fb/∂α_fb (α-approx is exact)
         end
 
         if bus_type_tb == PSY.ACBusTypes.PQ
-            Jv[idx_p_tb, idx_p_tb] += s.common_tap_i # ∂P_tb/∂V_tb
+            Jv[idx_p_tb, idx_p_tb] += dP_dV_tb # ∂P_tb/∂V_tb
             Jv[idx_q_tb, idx_p_tb] +=
                 _calculate_dQ_dV_lcc(s.tap_i, s.i_dc, xtr_i, Vm_tb, phi_i) # ∂Q_tb/∂V_tb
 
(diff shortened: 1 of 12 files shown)


src/lcc_utils.jl
```diff
@@ -34,8 +41,14 @@ function _calculate_dQ_dV_lcc(
...
-    return Vm * t * sqrt(6) / π * I_dc *
-           (sin(ϕ) - cos(ϕ) * x_t * I_dc / (sqrt(2) * Vm * t * sin(ϕ)^2))
+    sϕ = sin(ϕ)
+    # On the clamp boundary (sin(ϕ) = 0), φ is locally pinned and the residual
+    # is constant in this direction, so the true derivative is 0 even though
+    # the analytic formula has a 1/sin(ϕ) singularity.
+    sϕ < 1e-8 && return 0.0
+    return t * sqrt(6) / π * I_dc * sϕ -
+           sqrt(6) / π * cos(ϕ) * sign(I_dc) * I_dc^2 * x_t /
+           (sqrt(2) * Vm * sϕ)
```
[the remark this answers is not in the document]
luke-kiernan: This is numerically defensible: `raw` is within machine precision of hitting the bounary when `sin(ϕ) < 1e-8`.

The detailed math: for small `x`, `sqrt(1-x) ≈ 1 - x/2`.
`raw = cos(ϕ) = ±sqrt(1 - sin^2(ϕ)) ≈ ±(1 - 0.5 sin^2(ϕ))`.  `eps(Float64) ≈ 2e-16`, so we lose precision around`0.5 * sin^2(ϕ) ≈ 2e-16` or `sin(ϕ) ≈ e-8`

src/lcc_utils.jl
```diff
@@ -50,8 +63,11 @@ function _calculate_dQ_dt_lcc(
     Vm::Float64,
     ϕ::Float64,
 )::Float64
-    return Vm * t * sqrt(6) / π * I_dc *
-           (sin(ϕ) / t - cos(ϕ) * x_t * I_dc / (sqrt(2) * Vm * t^2 * sin(ϕ)^2))
+    sϕ = sin(ϕ)
+    sϕ < 1e-8 && return 0.0
+    return Vm * sqrt(6) / π * I_dc * sϕ -
```
luke-kiernan: nitpick: explicit `sqrt(6)/\pi` vs `sqrt_6_div_pi` constant. I tend to think we should just be uniform and go with the constant everywhere

jd-lara: sϕ < 1e-8 && return 0.0 -> should we define why this tolerance in a const?

luke-kiernan: ~~It's only used in a single spot.~~ But I can make it a constant for readability

edit: oh, it's in 3 spots. Then definitely worth replacing with global const

src/ac_power_flow_jacobian.jl
jd-lara, line 410: @luke-kiernan I believe these are now outdated with this change

jd-lara (changes requested):
LGTM but rebase on top of main and ask AI to find duplications please

src/ac_power_flow_jacobian.jl
luke-kiernan: What do you mean? Which change? This PR only corrects entries in the Jacobian, doesn't add new variables (so now new rows/columns).

luke-kiernan:
Rebased. One thing that came up in the code review and testing refactor: ~~the reason we're getting blow up near the boundary is in part because we assume `α ≈ ϕ`.~~ No, not quite right. The `α`-based formula is equivalent in the interior, but behaved worse at the boundary. I'll rewrite it in terms of `ϕ` and add tests for the boundary behavior.

src/ac_power_flow_jacobian.jl
```diff
@@ -545,67 +552,87 @@ function _set_entries_for_lcc(data::ACPowerFlowData,
...
+        s = _lcc_jacobian_scalars(data, i, time_step, Vm_fb, Vm_tb)
+
+        # True-ϕ ∂P/∂{V, t, α} entries come from `_lcc_jacobian_scalars`,
+        # which calls the `_calculate_dP_d{V,t,α}_lcc` helpers (boundary-
+        # guarded at `sin(ϕ) → 0`). In the interior these are algebraically
+        # equal to the legacy α-approximation form via the ϕ-definition
+        # identity `cos(α) - cos(ϕ) = x_t·I_dc/(√2·V·tap)`; at the clamp
+        # the helpers correctly drop the chain term (where `∂ϕ/∂x = 0`),
+        # while the α-form silently uses `cos(α) ≠ cos(ϕ_clamp)`.
+        # See `test_jacobian.jl::"Jacobian verification with LCC at inverter ϕ clamp"`.
```
luke-kiernan: AI loves its long-winded comments. Trim or eliminate: I don't see this explanation as adding something of value

src/rectangular_ci_power_flow_jacobian.jl
```diff
@@ -690,81 +693,104 @@ function _set_entries_for_lcc_rect!(
...
+        s = _lcc_jacobian_scalars(data, i, time_step, Vm_fb, Vm_tb)
+
+        # True-ϕ formulation throughout. Both LCC bus residual rows use the
+        # admittance form `F_real = -Re(Y·V) = -A·u(ϕ)`,
+        # `F_imag = -Im(Y·V) = -A·w(ϕ)` with `A = tap·√6/π·I_dc/V` and
+        # universal `u(ϕ) = cos(ϕ)·e + sin(ϕ)·f`, `w(ϕ) = cos(ϕ)·f − sin(ϕ)·e`.
+        # The inverter convention's sign flip (Y picks up an exp(-iπ) ≈ −1
+        # from ϕ_i ≈ π − α_i at x_t = 0) is absorbed into `cos(ϕ_i)`,
+        # `sin(ϕ_i)` — no separate sign handling needed in the Jacobian.
+        phi_r = data.lcc.rectifier.phi[i, time_step]
```
luke-kiernan, line 719: nitpick: this would likely be marginally cleaner if we worked in complex numbers...but probably not worth it. [The `u_fb, w_fb, u_tb, w_tb` expressions would clean up, but I don't spot other immediate simplifications.]

test/test_utils/jacobian_verification.jl
```diff
@@ -0,0 +1,119 @@
...
+)
+    n = length(x0)
+    residual(x0, time_step)
+    F0 = copy(residual.Rv)
+    F0_scale = max(LinearAlgebra.norm(F0), 1.0)
+    # Raw-remainder noise floor — below this, the remainder is dominated by
+    # floating-point roundoff from cancellation in `F(x+Δx) - F(x) - Δx·J·u`
+    # and the asymptotic ratio test is meaningless. The factor 1e4 accounts
+    # for typical accumulated relative roundoff in residuals built from many
+    # trig / sqrt / division operations (e.g. LCC residual).
```
luke-kiernan: how to make sure this filter isn't too lenient, attributing things to noise that aren't actually noise?

luke-kiernan (commented):
The test noise floor is the big one here. The other 2 comments I'd be fine as-is

src/lcc_utils.jl
```diff
@@ -163,6 +305,181 @@ function _update_ybus_lcc!(
...
+        dP_dt_fb = dP_dt_fb,
+        dP_dt_tb = dP_dt_tb,
+        dP_dα_fb = dP_dα_fb,
+        dP_dα_tb = dP_dα_tb,
+        # Tail-row × tail-column block (6 entries). F_t_fb has the
+        # P_lcc_from contribution (in the setpoint_at_rectifier case);
+        # F_t_tb has both P_lcc_from and P_lcc_to. d_Ft_fb_d_alpha_i is
+        # zero because F_t_fb doesn't depend on α_i.
+        d_Ft_fb_d_tap_r = dP_dt_fb,
+        d_Ft_fb_d_alpha_r = dP_dα_fb,
```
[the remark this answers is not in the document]
luke-kiernan: This PR has snowballed enough: I opened an issue for this and will address in a follow-up.

jd-lara (approved):

PR #367, opened 2026-05-15
Rectangular as Formulation 

jd-lara:
This is not that big of a change as it seems. It moves the rectangular code from solver to formulation

### DIFF
--- a/src/power_flow_method.jl
+++ b/src/power_flow_method.jl
@@ -521,6 +521,33 @@ function _iwamoto_step(time_step::Int,
     return true
 end
 
+# `validate_voltage_magnitudes` indexes the state as polar `[|V|, θ, …]`
+# (x[2i-1] = |V|). The rectangular CI state is `(e, f, …)` blocks, so the polar
+# check is meaningless there — dispatch a no-op for the rectangular residual
+# rather than gating with a runtime type check.
+function _validate_state_magnitudes(
+    ::ACPowerFlowResidual,
+    x::Vector{Float64},
+    bus_types::AbstractArray{PSY.ACBusTypes},
+    range::MinMax,
+    i::Int64,
+)
+    validate_voltage_magnitudes(x, bus_types, range, i)
+    return
+end
+
+function _validate_state_magnitudes(
+    ::ACRectangularCIResidual,
+    x::Vector{Float64},
+    bus_types::AbstractArray{PSY.ACBusTypes},
+    range::MinMax,
+    i::Int64,
+)
+    # TO BE IMPLEMENTED: add rectangular CI validation if needed. For now, just skip it since the rectangular CI solver doesn't use polar voltage magnitudes.
+    # We might need to do some 0 -1 validation for entries.
+    return
+end
+
 """Runs the full `NewtonRaphsonACPowerFlow`.
 # Keyword arguments:
 - `maxIterations::Int`: maximum iterations. Default: $DEFAULT_NR_MAX_ITER.
@@ -582,7 +609,8 @@ function _run_power_flow_method(time_step::Int,
                 refinement_eps,
             )
         end
-        validate_vms && PowerFlows.validate_voltage_magnitudes(
+        validate_vms && _validate_state_magnitudes(
+            residual,
             stateVector.x,
             bus_types,
             vm_validation_range,
@@ -660,7 +688,8 @@ function _run_power_flow_method(time_step::Int,
             autoscale,
             iwamoto_fallback,
         )
-        validate_vms && PowerFlows.validate_voltage_magnitudes(
+        validate_vms && _validate_state_magnitudes(
+            residual,
             stateVector.x,
             bus_types,
             vm_validation_range,
--- a/src/power_flow_method.jl
+++ b/src/power_flow_method.jl
@@ -700,8 +700,28 @@ function _finalize_power_flow(
     return false
 end
 
+"""Formulation-specific post-Newton step. Polar needs nothing; the rectangular
+CI formulation distributes the converged subnetwork slack into the bus
+injection arrays."""
+_finalize_formulation!(::ACPolarPowerFlow, data, x, residual, time_step) = nothing
+
+function _finalize_formulation!(
+    ::ACRectangularPowerFlow,
+    data::ACPowerFlowData,
+    x::Vector{Float64},
+    residual::ACRectangularCIResidual,
+    time_step::Int64,
+)
+    rect_finalize_bus_injections!(
+        data, x, residual.bus_state_offset, residual.P_net_set,
+        residual.bus_slack_participation_factors, residual.subnetworks,
+        time_step,
+    )
+    return
+end
+
 function _newton_power_flow(
-    pf::ACPowerFlow{T},
+    pf::AbstractACPowerFlow{T},
     data::ACPowerFlowData,
     time_step::Int64;
     # shared kwargs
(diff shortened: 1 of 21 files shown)


src/power_flow_types.jl
```diff
@@ -253,23 +236,146 @@ function ACPowerFlow{ACSolver}(;
...
+    enhanced_flat_start::Bool = true,
+    skip_redistribution::Bool = false,
+    distribute_slack_proportional_to_headroom::Bool = false,
+    network_reductions::Vector{PNM.NetworkReduction} = PNM.NetworkReduction[],
+    time_steps::Int = 1,
+    time_step_names::Vector{String} = String[],
+    correct_bustypes::Bool = false,
+    solver_settings::Dict{Symbol, Any} = Dict{Symbol, Any}(),
+) where {ACSolver <: ACPowerFlowSolverType}
+    if ACSolver <: Union{
```
luke-kiernan, line 340: nitpick: could be multiple-dispatched. Like simply
```julia
function ACRectangularPowerFlow{ACSolver}{; kwargs...) where
       {ACSolver <: Union{[disallowed types]}}
    throw(...)
end
```

src/rectangular_ci_setup.jl
```diff
@@ -54,42 +54,47 @@ function fold_zip_constant_z!(
...
 """
-    rect_initial_state!(x, data, bus_state_offset, bus_block_size, time_step)
+    _rect_fill_state!(x, data, bus_state_offset, type_time_step, value_time_step)
 
-Initialize the state vector `x` from `data.bus_magnitude`, `data.bus_angles`,
-and the bus power-injection fields, plus the LCC tap/angle fields. Counterpart
-of [`rect_update_data!`](@ref). At REF buses, the first two slots hold
-`(P_gen, Q_gen)` (including any distributed-slack increment); elsewhere
-the first two slots hold `(e, f)`, and PV buses' third slot holds `Q_gen`.
+Fill the rectangular state vector `x`. The state-block *layout* (which buses
```
luke-kiernan, line 59: Oh yeah, that's an interesting point: rectangular doesn't cooperate with PV-PQ bus type swaps very well, if at all. We should add a test for this.

test/test_rectangular_ci_setup.jl
```diff
@@ -102,3 +102,115 @@ end
...
+        pf_off = ACRectangularPowerFlow{NewtonRaphsonACPowerFlow}(;
+            enhanced_flat_start = false)
+        data2 = PF.PowerFlowData(pf_off, sys)
+        residual2 = PF.ACRectangularCIResidual(data2, 1)
+        x0_off = PF.improve_x0(pf_off, data2, residual2, 1)
+        @test length(x0_off) == length(residual2.Rv)
+        @test all(isfinite, x0_off)
+    end
+
+    @testset "solves to the same answer as before (regression)" begin
```
luke-kiernan: Hmm. "solves to the same answer as before" makes it sound like we ought to be comparing the solutions, not just checking that it converges.

test/test_power_flow_data.jl
```diff
@@ -191,3 +191,56 @@ end
...
+    @test PF.get_robust_power_flow(r) === false
+    @test PF.get_slack_participation_factors(r) === nothing
+    @test PF.get_time_steps(r) == 1
+    @test PF.get_enhanced_flat_start(r) === true
+    p = ACPolarPowerFlow{NewtonRaphsonACPowerFlow}(;
+        calculate_loss_factors = true)
+    @test PF.get_calculate_loss_factors(p) === true
+end
+
+@testset "ACPowerFlowData covers rectangular model" begin
```
luke-kiernan: This last testset seems pretty silly. I'd remove it

test/test_power_flow_data.jl
```diff
@@ -191,3 +191,56 @@ end
...
+@testset "ACPowerFlow alias resolves to ACPolarPowerFlow" begin
+    @test PF.ACPowerFlow === PF.ACPolarPowerFlow
+    @test PF.ACPolarPowerFlow <: PF.AbstractACPowerFlow
+    @test PF.AbstractACPowerFlow <: PF.PowerFlowEvaluationModel
+    pf = ACPowerFlow{NewtonRaphsonACPowerFlow}()
+    @test pf isa PF.ACPolarPowerFlow{NewtonRaphsonACPowerFlow}
+    @test ACPowerFlow() isa PF.ACPolarPowerFlow{NewtonRaphsonACPowerFlow}
+end
+
+@testset "ACRectangularPowerFlow type" begin
```
luke-kiernan, line 204: A little bit redundant--much of this is also covered in `test/test_rectangular_ci_power_flow.jl` [here](https://github.com/Sienna-Platform/PowerFlows.jl/pull/367/changes#r3251444037)

test/test_rectangular_ci_power_flow.jl
```diff
@@ -44,62 +44,26 @@ end
...
-    end
-
-    @testset "step_strategy=:typo rejected with helpful msg" begin
-        settings = merge(base_settings,
-            Dict{Symbol, Any}(:step_strategy => :trust_regioon))
-        pf = ACPowerFlow{RectangularCurrentInjectionACPowerFlow}(;
-            solver_settings = settings)
-        @test_throws ArgumentError solve_power_flow(pf, deepcopy(sys))
-    end
+@testset "Rectangular CI Power Flow: unsupported config rejected" begin
```
luke-kiernan, line 61: De-duplicate with [this](https://github.com/Sienna-Platform/PowerFlows.jl/pull/367/changes#r3251441794)

src/power_flow_method.jl
```diff
@@ -700,8 +700,28 @@ function _finalize_power_flow(
...
+_finalize_formulation!(::ACPolarPowerFlow, data, x, residual, time_step) = nothing
+
+function _finalize_formulation!(
+    ::ACRectangularPowerFlow,
+    data::ACPowerFlowData,
+    x::Vector{Float64},
+    residual::ACRectangularCIResidual,
+    time_step::Int64,
+)
+    rect_finalize_bus_injections!(
```
luke-kiernan, line 744: Do we have tests for rectangular with distributed slack? Also should test rectangular with headroom-proportional slack too.

luke-kiernan (commented):
To what degree should we aim for testing parity between rectangular and polar? Here I mean with regards to things like

- network reductions
- distributed slack (headroom-proportional, time-varying)
- convergence failure negative test (convergence=false code path)
- reactive power redistribution

Perhaps it's worthwhile for some but not for others. (I'd expect network reductions to be formulation-orthogonal. Probably reactive power redistribution too.) This could and maybe should be deferred to a later PR--my comments here are more about the rectangular method as a whole than the rename.

Oh also we should probably add a test somewhere that polar and rectangular converge to the same thing.

jd-lara:
> To what degree should we aim for testing parity between rectangular and polar? Here I mean with regards to things like
> 
> * network reductions
> * distributed slack (headroom-proportional, time-varying)
> * convergence failure negative test (convergence=false code path)
> * reactive power redistribution
> 
> Perhaps it's worthwhile for some but not for others. (I'd expect network reductions to be formulation-orthogonal. Probably reactive power redistribution too.) This could and maybe should be deferred to a later PR--my comments here are more about the rectangular method as a whole than the rename.
> 
> Oh also we should probably add a test somewhere that polar and rectangular converge to the same thing.

I think we need to get to parity on the pieces that make sense like losses, reductions, distributed slack etc.

luke-kiernan (approved):
Just realized I hit "comment" not "approve" (and you said you wanted to get this merged so you can wrap up the SCUC thing over the weekend)

src/power_flow_method.jl
jd-lara: added here [33c2f31](https://github.com/Sienna-Platform/PowerFlows.jl/pull/367/commits/33c2f312088a3760609317c41db23d777b8d53e1)

Issue #258, opened 2025-11-05
LCC result validation


luke-kiernan:
Some loose ends on adding LCC support:

1. Write more test cases for AC power flows with LCC elements: for example, a system with 2 LCCs. [I believe all the ones in our current test cases only have 1.]
2. Validate against PSS/E: figure out how that program models LCCs, in terms of design choices,  and imitate their methods as best as we can.
3. Add a test case where there's an LCC with nonzero loss.

Also, we might want to add a system featuring an LCC to PSB. [I'm not seeing any there, though it's possible I missed one...]

PR #372, opened 2026-05-17
hybrid cpb formulation

jd-lara:
@luke-kiernan after looking at the LM performance with the rectangular model I realized that there were structural weaknesses to LM with rectangular. 

Long story short, I added a mixed formulation that uses Rectangular in PQ buses and Polar on PV buses which performs really well with LM and the other solvers too. 

This PR also adds proper scaling optionality to LM to make it not as bad for rectangular and protections. 

I also asked Claude to document the formulation as well as make a script to benchmark and add notes on model and solver selector. 

overall is a long PR but a step forward on the search for improved algorithmic and moiling efficiencies in PF. 

This PR also fixes a broken test in main from #370


src/ac_power_flow_residual.jl
```diff
@@ -13,6 +13,7 @@ A struct to keep track of the residuals in the Newton-Raphson AC power flow calc
 - `bus_slack_participation_factors::SparseVector{Float64, Int}`: A sparse vector of the slack participation factors aggregated at the bus level.
 - `subnetworks::Dict{Int64, Vector{Int64}}`: The dictionary that identifies subnetworks (connected components), with the key defining the REF bus, values defining the corresponding buses in the subnetwork.
 - `P_slack_buf::Vector{Float64}`: Scratch buffer of length `n_buses` used by `_update_residual_values!` to write the per-subnetwork slack distribution in place, avoiding a per-iteration allocation when indexing `bus_slack_participation_factors` by `subnetwork_buses`.
+- `validate_indices::Vector{Int}`: precomputed `x`-indices of PQ-bus |V| entries for the per-iteration voltage-magnitude diagnostic.
```
luke-kiernan, line 16: Why use a list of indices when you can use a bitmask? First is random access pattern, second linear. Same goes in other spots where you're doing bus type based selection.

src/common.jl
```diff
@@ -560,35 +560,88 @@ function make_bus_slack_participation_factors!(
...
+          "$num_outside_range $bus_label buses." maxlog = PF_MAX_LOG
+    return
+end
+
+"""Precompute, once per solve, the `x`-offsets of PQ/PV buses for the per-bus
+block (rectangular CI / mixed CPB) state layout. Bus types and the state
+layout are invariant across NR/TR iterations, so this filtering is hoisted out
+of the per-iteration validator. `offsets[i]` is the start of bus `i`'s block;
+`(e, f) = (x[off], x[off + 1])`. REF is fixed and excluded here."""
+function _pqpv_validate_offsets(
```
luke-kiernan, line 582: Might be simpler to just skip REF buses. (We shouldn't have isolated, I believe...) The PNM y bus struct may have a pre-computed set of ref buses already.

src/levenberg-marquardt.jl
```diff
@@ -67,16 +80,44 @@ function copy_jacobian!(ws::LMWorkspace, Jv::SparseMatrixCSC{Float64, J_INDEX_TY
     return
 end
 
-"""Update the √λ diagonal and re-factorize."""
+"""Running max of the J column 2-norms (MINPACK lmder, mode 1). A zero column
```
luke-kiernan: > MINPACK lmder, mode 1

what?

src/mixed_cpb_power_flow_jacobian.jl
```diff
@@ -0,0 +1,805 @@
...
+    Yvals = SparseArrays.nonzeros(Y_bus_eff)
+    offdiag_pv_i = Int[]
+    offdiag_pv_k = Int[]
+    offdiag_pv_y = ComplexF64[]
+    cols_nz = Vector{Int}[]
+    @inbounds for col in 1:n_buses
+        # REF columns hold (P_gen, Q_gen): no (e, f) off-diagonal entries exist.
+        bus_types[col] == PSY.ACBusTypes.REF && continue
+        col_off = Int(bus_state_offset[col])
+        for j in Y_bus_eff.colptr[col]:(Y_bus_eff.colptr[col + 1] - 1)
```
luke-kiernan: nitpick: this colptr range is just `nzrange(Y_bus_eff, col)`. Elsewhere too: search for instances of `colptr`.

src/mixed_cpb_power_flow_residual.jl
```diff
@@ -0,0 +1,316 @@
...
+    Y_bus_eff::SparseMatrixCSC{ComplexF64, Int}
+    P_net_const::Vector{Float64}
+    Q_net_const::Vector{Float64}
+    const_I_P::Vector{Float64}
+    const_I_Q::Vector{Float64}
+    P_net_set::Vector{Float64}
+    bus_slack_participation_factors::SparseVector{Float64, Int}
+    subnetworks::Dict{Int64, Vector{Int64}}
+    bus_state_offset::Vector{REC_INDEX_TYPE}
+    bus_block_size::Vector{Int8}
```
luke-kiernan, line 28: > every bus occupies exactly 2 state entries

Then we don't need these fields and should just hard-code the values (block size 2, offsets `2i-1`)

test/test_mixed_cpb_jacobian.jl
```diff
@@ -0,0 +1,320 @@
+@testset "Mixed CPB Jacobian: structure + constant PQ off-diagonal blocks" begin
+    function _check_mixed_cpb_jacobian(sys, label)
```
luke-kiernan: Looks like this test case was written based on the code, not based on its intended behavior. As AI tends to do.

test/test_mixed_cpb_jacobian.jl
```diff
@@ -0,0 +1,320 @@
...
+        return R, x
+    end
+
+    @testset "c_sys5" begin
+        sys = PSB.build_system(PSB.PSITestSystems, "c_sys5")
+        R, x = _build_mixed_x(sys)
+        Random.seed!(2024)
+        x .+= 1e-3 .* randn(length(x))
+        err = _fd_check(R, x)
+        @test err < 1e-6
```
luke-kiernan: Should really be verifying asymptotic behavior, not matching to a fixed tolerance. I made a `verify_jacobian` function for this exact purpose: switch to using that.

src/mixed_cpb_power_flow_jacobian.jl
```diff
@@ -0,0 +1,805 @@
...
+- Slack cross-term caches `slack_nz_idx_e`/`slack_nz_idx_f` + `slack_c_k`
+- LCC tail nzval cache `lcc_nz` (20×n_lccs)
+"""
+struct ACMixedCPBJacobian
+    data::ACPowerFlowData
+    Jf!::Function
+    Jv::SparseMatrixCSC{Float64, J_INDEX_TYPE}
+    Y_bus_eff::SparseMatrixCSC{ComplexF64, Int}
+    Y_diag::Vector{ComplexF64}     # cached Y_bus_eff diagonal; avoids O(log nnz) sparse access per iteration
+    e_state::Vector{Float64}       # shared view into residual's e_state
```
luke-kiernan, line 19: Yikes. 8 different shared views into residual. That's enough that I'd consider making a dedicated struct to hold them.

luke-kiernan (changes requested):
First pass. Big 2 comments:

1. Redundant offset/block size logic for the CBP method.
2. Test correctness of the CBP Jacobian by looking at asymptotic agreement with finite differences to the expected order.

Fix those and I'll take another look.

luke-kiernan:
```julia
    # uniformly 2; vector kept for signature parity with the rectangular formulation
    fill!(block_sizes, Int8(2))
    pos = REC_INDEX_TYPE(1)
    for i in 1:n_buses
        offsets[i] = pos
        pos += REC_INDEX_TYPE(2)
    end
```
Oh, so maybe leaving the offsets/block sizes was purposeful? I question the wisdom of this: as-is, I'd be pleasantly surprised if these constants were propagated by the compiler to their fullest extent. 

If the goal is inter-operability (being able to write code that works for both CBP and CI), then just make getters and multiple-dispatch: `get_block_size` returns `block_size[i]` for rectangular CI, `2` for CBP.

src/levenberg-marquardt.jl
jd-lara: It is the idea implemented in MINPACKS LM https://www.netlib.org/minpack/lmder.f 

This has a huge impact on the performance of LM between Polar and Rectangular

src/ac_power_flow_residual.jl
jd-lara: Good catch, I'll make the change

src/mixed_cpb_power_flow_jacobian.jl
jd-lara: lets improve this in a next iteration

src/ac_power_flow_residual.jl
jd-lara: I am going to not do this since it seems to cause a regression. PQ buses are majority in the test cases so it doesn't help as much

src/mixed_cpb_power_flow_residual.jl
jd-lara: deferred this to keep testing consistency with rectangular

src/ac_power_flow_residual.jl
luke-kiernan: > PQ buses are majority in the test cases so it doesn't help as much

Wait, what? I'd expect the opposite. The bitmask should win when there's many entries, the sparse vector when there's few.

src/mixed_cpb_power_flow_jacobian.jl
```diff
@@ -0,0 +1,733 @@
...
+    Ir_acc::Vector{Float64}        # shared view into residual's Ir_acc (accumulated Re(I) per bus)
+    Ii_acc::Vector{Float64}        # shared view into residual's Ii_acc (accumulated Im(I) per bus)
+    bus_slack_participation_factors::SparseVector{Float64, Int}
+    subnetworks::Dict{Int64, Vector{Int64}}
+    bus_state_offset::Vector{REC_INDEX_TYPE}
+    bus_block_size::Vector{Int8}
+    total_bus_state::Int
+    # nzval-index caches (populated once at construction)
+    diag_base_nz::Matrix{Int}        # 4 × n_buses; rows: (off,off), (off,off+1), (off+1,off), (off+1,off+1)
+    offdiag_pv_nz::Matrix{Int}       # 2 × n_pv_pairs; PV power-balance row only (voltage-constraint row has no off-diagonals); see "offdiag_pv_nz layout"
```
luke-kiernan: > see "offdiag_pv_nz layout"

Where? Not seeing anything with control + F.

src/power_flow_setup.jl
```diff
@@ -81,10 +81,52 @@ function improve_x0(pf::ACRectangularPowerFlow,
...
+rectangular path verbatim, swapping `rect_initial_state!`→`mixed_initial_state!`,
+`_rect_fill_state!`→`_mixed_fill_state!`, and the rectangular
+`_enhanced_flat_start`→the MCPB `ACMixedCPBResidual` overload. No DC robust
+fallback: `ACMixedPowerFlow` has no `robust_power_flow` field
+(`get_robust_power_flow(::AbstractACPowerFlow) == false`)."""
+function improve_x0(pf::ACMixedPowerFlow,
+    data::ACPowerFlowData,
+    residual::ACMixedCPBResidual,
+    time_step::Int64,
+)
```
luke-kiernan, line 97: Add the `x0 =` key word argument, or otherwise make it so that we can override `improve_x0` and specify an arbitrary starting point for testing purposes.

test/test_mixed_cpb_jacobian.jl
```diff
@@ -0,0 +1,228 @@
...
+        @test PF.solve_and_store_power_flow!(pf_polar, sys)
+        pf_mixed = ACMixedPowerFlow{NewtonRaphsonACPowerFlow}(;
+            correct_bustypes = correct_bustypes, solver_settings = settings)
+        data = PF.PowerFlowData(pf_mixed, sys)
+        R = PF.ACMixedCPBResidual(data, 1)
+        x = Vector{Float64}(undef, length(R.Rv))
+        PF.mixed_initial_state!(x, data, R.bus_state_offset, R.bus_block_size, 1)
+        return R, x
+    end
+
```
luke-kiernan, line 141: Are these 2 tests sufficient to check all the different potentially nonzero LCC-related Jacobian entries? The simple lcc system has just 3 buses...

test/test_mixed_cpb_jacobian.jl
```diff
@@ -0,0 +1,228 @@
...
+@testset "Mixed CPB solver-driver wiring" begin
+    sys = PSB.build_system(PSB.PSITestSystems, "c_sys5")
+    data = PF.PowerFlowData(ACMixedPowerFlow{NewtonRaphsonACPowerFlow}(), sys)
+    R, J, x0 = PF.initialize_power_flow_variables(
+        ACMixedPowerFlow{NewtonRaphsonACPowerFlow}(), data, 1)
+    @test all(isfinite, x0)
+
+    # Smoke check: the shared step functions accept the mixed R/J without a
+    # MethodError (full parity is Stage 3 Task 3.1).
```
luke-kiernan: > Stage 3 Task 3.1

Looks like instructions to Claude snuck in here. Elsewhere too.

luke-kiernan (changes requested):
The one thing I'd like fixed here is the `x0 =` keyword argument, for overriding the "improve initial guess" stuff.

src/common.jl
jd-lara: there are never many ref buses, this is kinda over engineering

test/test_mixed_cpb_jacobian.jl
jd-lara: we need to make more testing but we have #258 opened for that.

src/power_flow_setup.jl
jd-lara: done

test/test_mixed_cpb_jacobian.jl
jd-lara: I did another to reduce AI stupidity

src/mixed_cpb_power_flow_jacobian.jl
jd-lara: Dropped the dangling reference from the first pass implementation

src/levenberg-marquardt.jl
```diff
@@ -5,15 +5,13 @@ to update its entries in-place, and a cached QR factorization. `D` is the
...
-    # PERF: if we're doing dense access, I'd expect a bitmask to be faster.
-    #       Or only store λ_diag_indices (sparse) and infer J entries as the complement.
     # Indices into A.nzval for the J block entries (same order as J.Jv.nzval)
     j_nzval_indices::Vector{Int}
     # Indices into A.nzval for the √λ diagonal entries (length n)
     λ_diag_indices::Vector{Int}
-    # PERF: have not investigated other factorization methods. SPQR does not allow
-    #       for in-place updates or re-use of symbolic factorization, but with non
-    #       square matrices--J^T J form is more unstable--we don't have a lot of options.
+    # SPQR: enables a cached symbolic factorization with in-place numeric
```
luke-kiernan, line 12: Wait, you figured out a way to reuse the symbolic factorization?! When I asked Claude about that a while back, it told me that SPQR simply didn't allow for separate symbolic vs numeric factorization.

Edit: Oh. We're recomputing the whole thing--`ws.F = LinearAlgebra.qr(ws.A)`--and this comment is just wrong.

luke-kiernan (approved):
One misleading comment, but that's it.

PR #377, opened 2026-05-27
solver refactor

jd-lara:
opening as a draft to test in the CI-CD the Pardiso Backend.


ext/PowerFlowsPardisoExt.jl
```diff
@@ -0,0 +1,108 @@
+module PowerFlowsPardisoExt
+
+# MKLPardiso linear-solver backend for PowerFlows, loaded when `Pardiso.jl` is
+# available. Implements the PowerFlows-local solver-cache operations for
+# `PowerFlows.PardisoLinSolveCache` (declared in `src/linear_solver_backend.jl`).
+#
+# MKL Pardiso requires a functional MKL (x86_64 Linux/Windows); it is unavailable
```
luke-kiernan, line 7: nitpick: what about those old Intel Macs?

ext/PowerFlowsPardisoExt.jl
```diff
@@ -0,0 +1,108 @@
...
+    ps = Pardiso.MKLPardisoSolver()
+    _init_pardiso!(ps)
+    cache = PardisoLinSolveCache(ps, A, false)
+    finalizer(cache) do c
+        try
+            Pardiso.set_phase!(c.ps, Pardiso.RELEASE_ALL)
+            Pardiso.pardiso(c.ps)
+        catch
+        end
+    end
```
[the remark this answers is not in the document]
luke-kiernan: seconded: seems overly broad to swallow anything and everything.

src/definitions.jl
```diff
@@ -52,8 +52,11 @@ const BUS_VOLTAGE_MAGNITUDE_CUTOFF_MIN = 0.8
 const BUS_VOLTAGE_MAGNITUDE_CUTOFF_MAX = 1.2
 
 const TIs = Union{Int32, Int64}
-const J_INDEX_TYPE = Int32
-const REC_INDEX_TYPE = Int32
+# Int64 on Apple so AppleAccelerate's libSparse (Int64-only `columnStarts` ABI)
+# can factor the AC Jacobian natively; Int32 elsewhere (KLU only).
+const INDEX_TYPE = Sys.isapple() ? Int64 : Int32
```
luke-kiernan: Ensure that this constant is propogated at compile time: add `@static` in front and/or look at `code_lowered`

src/power_flow_method.jl
```diff
@@ -60,45 +63,63 @@ function _do_refinement!(stateVector::StateVectorCache,
...
-        else
-            @error("KLU factorization failed: $e")
-        end
-    else
+        # KLU signals a singular factorization by throwing; AppleAccelerate and
+        # MKLPardiso do not (see `_set_Δx_nr!` residual guard below). Any factorization
+        # error routes to the regularized fallback.
+        e isa LinearAlgebra.SingularException ||
+            @error("Linear solver factorization failed: $e")
+        use_fallback = true
```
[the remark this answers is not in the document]
luke-kiernan: seconded

src/linear_solver_backend.jl
```diff
@@ -0,0 +1,105 @@
...
+
+"""Cache for the MKLPardiso backend. The Pardiso handle is held opaquely because
+`Pardiso.MKLPardisoSolver`'s type is only available when the `Pardiso.jl`
+extension (`PowerFlowsPardisoExt`) is loaded; the operation methods live there.
+The matrix is snapshotted because Pardiso reads it at solve time. `Ti` is left
+abstract: Pardiso.jl converts the index arrays to `Int32` internally, so this
+cache accepts both the `Int32` AC Jacobian and the `Int`-indexed DC matrices."""
+mutable struct PardisoLinSolveCache
+    ps::Any                       # Pardiso.MKLPardisoSolver
+    A::SparseMatrixCSC{Float64}
```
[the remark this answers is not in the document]
luke-kiernan: Seconded: seems worth checking.

Attaching the element type as a type parameter could make vector math with the stored matrix specialize better. You could even add an abstract supertype `LinearSolverCache{T}` (where `T` is the element type), so that the vector math would specialize without knowing the factorization type (AA, KLU, etc.)

jd-lara:
Improvements thanks to the pre-allocations in DC PF for multi period 

image

ext/PowerFlowsPardisoExt.jl
```diff
@@ -0,0 +1,125 @@
...
+    Pardiso.pardiso(cache.ps, cache.scratch, cache.A, b)
+    copyto!(b, cache.scratch)
+    return b
+end
+
+function PowerFlows.solve!(cache::PardisoLinSolveCache, b::StridedMatrix{Float64})
+    Pardiso.set_phase!(cache.ps, Pardiso.SOLVE_ITERATIVE_REFINE)
+    # Reuse a persistent buffer so a loop of multi-RHS solves on a cached factorization
+    # (e.g. multi-period / PCM DC) does not allocate an n×nrhs matrix per call. The shape
+    # is stable across such a loop, so the resize happens at most once.
```
luke-kiernan, line 104: We don't loop for multi-period DC. We handle all time periods via a single multi-column `Ax = B` solve. Methodology could still make sense if this gets called multiple times in a loop, but comment is confusing.

edit: okay I see now. You mean the outer "multi timestep PCM, each with DC solve" loop, not the inner "multi-timestep DC solve" loop

ext/PowerFlowsPardisoExt.jl
```diff
@@ -0,0 +1,108 @@
...
+function PowerFlows.solve_w_refinement(
+    cache::PardisoLinSolveCache,
+    A::SparseMatrixCSC{Float64},
+    b::Vector{Float64},
+    refinement_eps::Float64,
+)
+    x = copy(b)
+    PowerFlows.solve!(cache, x)
+    return x
+end
```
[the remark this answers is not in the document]
luke-kiernan: Agreed: it's confusing that

1. `A` is completely ignored.
2. This is identical to a `solve!` call.

src/ac_power_flow_jacobian.jl
```diff
@@ -872,7 +875,10 @@ function _singular_value_decomposition(
     tol::Float64 = 1e-9,
     max_iter::Integer = 100,
 )
-    factorized_block_J = KLU.klu(Jv)
+    # Voltage-stability factors require transpose solves (Aᵀ \ b), which only the
+    # KLU backend provides; this routine is KLU-only by construction.
```
luke-kiernan: Confusing comment. It's support for solving `A^T x = b` _using a factorization of A_. "Transpose solves" sounds like just plain solving `A^T x = b`, which is standard.

(I could check if AA supports this, but imo probably not worth it. Not in a performance critical path.)

src/power_flow_method.jl
```diff
@@ -23,6 +23,11 @@ struct StateVectorCache
     Δx_cauchy::Vector{Float64} # Cauchy step
     Δx_nr::Vector{Float64} # Newton-Raphson step
     d::Vector{Float64}
+    # Persistent KLU cache for the regularized singular-Jacobian fallback. Lazily built on
+    # the first fallback and reused via `numeric_refactor!` while the regularized matrix's
+    # sparsity pattern is unchanged; rebuilt with a full factor when the pattern shifts
+    # (it can, e.g. near a flat start where some Jacobian entries vanish/reappear).
+    fallback_cache::Base.RefValue{Union{Nothing, PNM.KLULinSolveCache{Float64}}}
```
luke-kiernan: Does this addition have measurable performance benefits, or is it mostly precautionary?

src/power_flow_method.jl
```diff
@@ -60,45 +71,81 @@ function _do_refinement!(stateVector::StateVectorCache,
...
+        # reuse is gated by `numeric_refactor!`'s own pattern check — which knows the
+        # cache's internal (0-based) layout — so we let it throw on a mismatch and rebuild,
+        # rather than re-deriving that comparison here.
+        cache = stateVector.fallback_cache[]
+        reused = false
+        if cache !== nothing
+            try
+                numeric_refactor!(cache, M)
+                reused = true
+            catch
```
luke-kiernan: Overly broad `catch`. Also, the symbolic refactor avoided if you switch `_singular_J_fallback` to use a structure-preserving `J' * J`: see `A_plus_eq_BT_B!` in homotopy_hessian.jl.

src/power_flow_method.jl
```diff
@@ -60,45 +71,81 @@ function _do_refinement!(stateVector::StateVectorCache,
...
-        else
-            @error("KLU factorization failed: $e")
-        end
-    else
+        # KLU signals a singular factorization by throwing; AppleAccelerate and
+        # MKLPardiso do not (see `_set_Δx_nr!` residual guard below). Any factorization
+        # error routes to the regularized fallback.
+        e isa LinearAlgebra.SingularException ||
+            @error("Linear solver factorization failed: $e")
+        use_fallback = true
```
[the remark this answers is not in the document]
luke-kiernan: Seconded

src/solve_dc_power_flow.jl
```diff
@@ -153,13 +220,19 @@ function solve_power_flow!(
...
-        data.arc_active_power_flow_to_from .= -data.arc_active_power_flow_from_to
-        Rs = _get_arc_resistances(data)
-        data.arc_active_power_losses .= Rs .* data.arc_active_power_flow_from_to .^ 2
+        mul!(
+            data.arc_active_power_flow_from_to,
+            transpose(data.aux_network_matrix.data),
+            data.bus_angles,
+        )
+        @. data.arc_active_power_flow_to_from = -data.arc_active_power_flow_from_to
+        data.arc_active_power_losses .=
```
luke-kiernan: Is there any logic to why you use `@.` in some spots here, and broadcast individual operations in others?

src/solve_dc_power_flow.jl
```diff
@@ -19,6 +19,59 @@ function adjust_power_injections_for_lccs!(power_injections::Matrix{Float64},
...
+    p_inj_dims = size(view(data.bus_active_power_injections, valid_ix, :))
+    return (
+        power_injections = similar(data.bus_active_power_injections),
+        p_inj = Matrix{Float64}(undef, p_inj_dims),
+        rs = _get_arc_resistances(data),
+        arc_bus_incidence = _build_signed_arc_bus_incidence(data),
+    )
+end
+
+function _build_signed_arc_bus_incidence(data::PowerFlowData)
```
luke-kiernan: Organizationally, this would belong better in PNM. 

Also, you're using element type `Float64` when all entries are 1/-1. Purposeful?

src/solve_dc_power_flow.jl
```diff
@@ -153,13 +220,19 @@ function solve_power_flow!(
         data.arc_active_power_losses .=
             data.arc_active_power_flow_from_to .+ data.arc_active_power_flow_to_from
     else
-        data.arc_active_power_flow_from_to .=
-            data.aux_network_matrix.data' * data.bus_angles
-        data.arc_active_power_flow_to_from .= -data.arc_active_power_flow_from_to
-        Rs = _get_arc_resistances(data)
-        data.arc_active_power_losses .= Rs .* data.arc_active_power_flow_from_to .^ 2
+        mul!(
```
luke-kiernan, line 209: I suspect you could fuse the `transpose` here with the `mul!`, with the right linear algebra function.

src/solve_dc_power_flow.jl
```diff
@@ -32,25 +85,30 @@ This function modifies the following fields of `data`, setting them to the compu
...
+    p_inj = scratch.p_inj
+    @views p_inj .= power_injections[valid_ix, :]
     solve!(solver_cache, p_inj)
-    data.bus_angles[valid_ix, :] .= p_inj
-    _compute_arc_angle_differences_from_data!(data)
-    Rs = _get_arc_resistances(data)
-    data.arc_active_power_losses .= Rs .* data.arc_active_power_flow_from_to .^ 2
+    @views data.bus_angles[valid_ix, :] .= p_inj
+    mul!(data.arc_angle_differences, scratch.arc_bus_incidence, data.bus_angles)
+    data.arc_active_power_losses .= scratch.rs .* data.arc_active_power_flow_from_to .^ 2
```
luke-kiernan: Confusing: where do we write to `scratch.rs`?

src/linear_solver_backend.jl
```diff
@@ -0,0 +1,107 @@
...
+# via dispatch over the PFLinearSolverCache Union. A future PNM-side abstract
+# supertype will let us drop this Union.
+
+"""Cache for the MKLPardiso backend. The Pardiso handle is held opaquely because
+`Pardiso.MKLPardisoSolver`'s type is only available when the `Pardiso.jl`
+extension (`PowerFlowsPardisoExt`) is loaded; the operation methods live there.
+The matrix is snapshotted because Pardiso reads it at solve time. `Ti` is left
+abstract: Pardiso.jl converts the index arrays to `Int32` internally, so this
+cache accepts both the `Int32` AC Jacobian and the `Int`-indexed DC matrices."""
+mutable struct PardisoLinSolveCache
```
luke-kiernan, line 21: Does this need to be a mutable struct? performance reasons

luke-kiernan (changes requested):
First pass. 

1. About `PFLinearSolverCache`: Is this compiler-friendly? Type-stable, inlining, constant proagation, etc. One concrete suggestion: look into how BLAS backend switching works and imitate their strategy at a high level. The fact that the MKL wrapper ends up here instead of in PNM is also organizationally strange. It'd be simpler if all the types had a common abstract supertype and a common interface.
2. The changes to the DC solve: organizationally confusing. We have `@views` and reads from `data.solver_cache`, `@.` and individually broadcasted operations. Why do some things end up under `solver_cache` whereas others end up under `data`? What purpose is `Base.Ref` serving here? Also, confusing name choices: I'd expect that `solver_cache` is of type `PFLinearSolverCache`, but nope.
3. Storing the `fallback_cache`: the symbolic refactor avoided if you switch `_singular_J_fallback` to use a structure-preserving `J' * J`: see `A_plus_eq_BT_B!` in homotopy_hessian.jl. Also uses `Base.Ref`, which makes me wonder if it's over-engineered.

ext/PowerFlowsPardisoExt.jl
luke-kiernan: I'll resolve this. Bigger fish to fry

src/linear_solver_backend.jl
jd-lara: Yes to overwrite the pointer later

PR #380, opened 2026-05-29
Bump version to 0.19.0 and update dependencies

jd-lara:
Updated project version and dependencies.


jd-lara (commented):
@copilot update the test/project.toml to use v 0.22 of PowerNetworkMatrices and remove the sources entry

jd-lara:
@copilot fix the performance test

PR #381, opened 2026-05-29
Reactive power control loops 

### LINKED ISSUE #127: Update the effects of the Facts devices
Include the effect of the Facts devices as static reactive power injections or something static that makes sense I think constant negative reactive power load and throw errors if the control mode isn’t supported.

jd-lara:
This PR implements the homotopy approach Agarwal's paper and also improvements to handle the outer loop reactive power limits check. 

image

### DIFF
--- a/src/ac_power_flow_jacobian.jl
+++ b/src/ac_power_flow_jacobian.jl
@@ -115,6 +115,18 @@ J.Jv  # Access the Jacobian matrix stored internally in J.
 # `ACPowerFlowJacobian` owns a fresh mutable buffer (the Newton loop mutates it in place;
 # Q-limit/PCM hold several live instances). Every nzval entry is rewritten by
 # `J(time_step)`, so the cached template stays pristine.
+# Reuse the cached structure on a matching key (network-matrix identity + slack nonzero pattern),
+# else `nothing` to signal a rebuild. The cache lives in its own `data.ac_jacobian_structure_cache`
+# field ([`ACJacobianStructureCache`](@ref)), so it never collides with the FastDecoupled/DC caches
+# in `data.solver_cache[]` (both can be live in a single FD-handoff-to-NR solve).
+_reuse_ac_jac_structure(::Nothing, matrix, nzind) = nothing
+_reuse_ac_jac_structure(e::ACJacobianStructureCache, matrix, nzind) =
+    if e.matrix === matrix && e.nzind == nzind
+        copy(e.structure)
+    else
+        nothing
+    end
+
 function _get_or_build_jacobian_structure(
     data::ACPowerFlowData,
     slack_factors::SparseVector{Float64, Int},
@@ -122,17 +134,13 @@ function _get_or_build_jacobian_structure(
     time_step::Int64,
 )
     nzind = SparseArrays.nonzeroinds(slack_factors)
-    entry = data.solver_cache[]
-    if entry isa Tuple{Symbol, Any, Vector{Int}, SparseMatrixCSC{Float64, J_INDEX_TYPE}} &&
-       entry[1] === :ac_jac_structure &&
-       entry[2] === data.power_network_matrix &&
-       entry[3] == nzind
-        return copy(entry[4])
-    end
+    reused = _reuse_ac_jac_structure(
+        data.ac_jacobian_structure_cache[], data.power_network_matrix, nzind)
+    isnothing(reused) || return reused
     Jv0 = _create_jacobian_matrix_structure(data, slack_factors, subnetworks, time_step)
     # Cache a pristine copy; `Jv0` is about to be mutated by the Newton loop.
-    data.solver_cache[] =
-        (:ac_jac_structure, data.power_network_matrix, copy(nzind), copy(Jv0))
+    data.ac_jacobian_structure_cache[] =
+        ACJacobianStructureCache(data.power_network_matrix, copy(nzind), copy(Jv0))
     return Jv0
 end
 
--- a/src/ac_power_flow_jacobian.jl
+++ b/src/ac_power_flow_jacobian.jl
@@ -8,16 +8,14 @@ and can be called as a function at the same time. Calling the instance as a func
 
 # Fields
 - `data::ACPowerFlowData`: The grid model data used for power flow calculations.
-- `Jf!::Function`: A function that calculates the Jacobian matrix inplace.
-- `Jv::SparseArrays.SparseMatrixCSC{Float64, $J_INDEX_TYPE}`: The Jacobian matrix, which is updated by the function `Jf!`.
+- `Jv::SparseArrays.SparseMatrixCSC{Float64, $J_INDEX_TYPE}`: The Jacobian matrix, which is updated by `_update_jacobian_matrix_values!`.
 - `diag_elements::MVector{4, Float64}`: Temporary storage for diagonal elements during Jacobian update.
 - `bus_slack_participation_factors::SparseVector{Float64, Int}`: Normalized per-bus slack participation factors for the current time step (from the `ACPowerFlowResidual`). Used for the distributed slack Jacobian entries.
 - `subnetworks::Dict{Int64, Vector{Int64}}`: Subnetwork mapping from REF bus to bus list (from the `ACPowerFlowResidual`). Used for the distributed slack Jacobian entries.
 """
 struct ACPowerFlowJacobian
     data::ACPowerFlowData
-    Jf!::Function   # This is the function that calculates the Jacobian matrix and updates Jv inplace
-    Jv::SparseArrays.SparseMatrixCSC{Float64, J_INDEX_TYPE}  # This is the Jacobian matrix, that is updated by the function Jf
+    Jv::SparseArrays.SparseMatrixCSC{Float64, J_INDEX_TYPE}  # This is the Jacobian matrix, updated in place by `_update_jacobian_matrix_values!`
     diag_elements::MVector{4, Float64}  # Temporary storage for diagonal elements during Jacobian update
     bus_slack_participation_factors::SparseVector{Float64, Int}
     subnetworks::Dict{Int64, Vector{Int64}}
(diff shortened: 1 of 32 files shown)


src/discrete_control/controlled_devices.jl
```diff
@@ -0,0 +1,184 @@
+abstract type AbstractControlledDevice end
+abstract type AbstractBranchControl <: AbstractControlledDevice end
+abstract type AbstractShuntControl <: AbstractControlledDevice end
+
+"""Voltage-controlling tap transformer. `nz_offsets` are the 4 cached
+`nzval` linear indices of the (from,to)×(from,to) Y-bus block."""
+mutable struct ControlledTap <: AbstractBranchControl
```
luke-kiernan, line 9: Do these really need to be mutable?

src/discrete_control/controlled_devices.jl
```diff
@@ -0,0 +1,184 @@
...
+parameter_limits(d::Union{ControlledPhaseShifter, ControlledFACTS}) = _seam_err(d)
+current_parameter(d::Union{ControlledPhaseShifter, ControlledFACTS}) = _seam_err(d)
+set_current_parameter!(d::Union{ControlledPhaseShifter, ControlledFACTS}, ::Float64) =
+    _seam_err(d)
+
+function _nz_index(A::SparseArrays.SparseMatrixCSC, row::Int, col::Int)
+    @inbounds for k in SparseArrays.nzrange(A, col)
+        A.rowval[k] == row && return k
+    end
+    error("Ybus has no stored entry at ($row,$col); structural zero")
```
luke-kiernan, line 196: nitpick: potentially brittle. iirc `SparseArrays.jl` does not guarantee that numerical zeros are preserved. But (1) there's no return value here that makes sense and (2) we implicitly rely on that elsewhere too, so it's fine.

src/ac_power_flow_jacobian.jl
```diff
@@ -107,19 +105,48 @@ J(time_step)  # Updates the Jacobian matrix stored internally in J.
...
 """
+# Memoize the expensive Jacobian sparse-structure build (~3.2 MB on 2000 buses) in the
+# persistent `data.solver_cache` Ref so it is built once and reused across the Q-limit
+# inner loop and repeated PCM solves. The structure depends only on network topology, REF
+# layout, and slack-participation pattern — none change on the PV→PQ flips that drive the
+# Q-limit loop (colptr/rowval verified byte-identical across a flip). Cache key is the
+# network-matrix identity + slack nonzero pattern, so a distributed-slack participant drop
+# (which changes the pattern) correctly rebuilds. A full `copy` is returned because some
+# nzval entries (e.g. LCC angle-constraint diagonals = 1.0) are set at structure creation
+# and never rewritten by `J(time_step)`, so they must be carried over, not zeroed.
```
luke-kiernan: If it's straightforward to write those values each iteration, I'd be in favor of doing so, for consistency reasons.

src/mixed_cpb_power_flow_jacobian.jl
```diff
@@ -12,7 +12,6 @@ hot path is `O(N + n_LCC)`. Field roles are in the inline comments below.
 """
 struct ACMixedCPBJacobian
     data::ACPowerFlowData
-    Jf!::Function
```
luke-kiernan, line 15: Intentional or AI-driven? I thought there were some performance benefits of making the function a field and/or a type parameter...same goes for the residual struct.

src/power_flow_method.jl
```diff
@@ -338,9 +342,14 @@ function _trust_region_step(time_step::Int,
     # Ratio of actual to predicted reduction
     LinearAlgebra.mul!(stateVector.r_predict, J.Jv, stateVector.Δx_proposed)
     stateVector.r_predict .+= stateVector.r
-    rho =
-        (sum(abs2, stateVector.r) - sum(abs2, residual.Rv)) /
-        (sum(abs2, stateVector.r) - sum(abs2, stateVector.r_predict))
+    predicted_reduction = old_residual_norm - sum(abs2, stateVector.r_predict)
+    # Non-positive predicted reduction (e.g. a non-descent fallback step) would give a
```
luke-kiernan: Should non-descent fallback steps happen? I'd think not, in which case I'd be inclined to error or at least `@warn`...

luke-kiernan (commented):
First pass. Will take a closer look at the details of the method and CoPilot's comments tomorrow

src/mixed_cpb_power_flow_jacobian.jl
jd-lara: Claude suggested it to avoid a type instability, I'll take a look in more depth

rodrigomha:
Should this change if we have discrete_control for SwitchedShunts:

https://github.com/Sienna-Platform/PowerFlows.jl/blob/03d240a8ecede1822c37192af5217adf0a890c95/src/common.jl#L128

It seems that we are adding to the injections directly but we want to handle it differently?

src/discrete_control/control_continuation.jl
```diff
@@ -0,0 +1,276 @@
...
+        trial = min(1.0, done + step)
+        p = start + trial * (target - start)
+        apply_parameter!(d, data, p, ts)
+        if _solve_with_q_limits!(pf, data, ts; kwargs...)
+            done = trial
+            reached = p
+            step = min(step * CONTROL_STEP_GROWTH, MAX_LAMBDA_STEP)
+        else
+            apply_parameter!(d, data, reached, ts)
+            step /= 2.0
```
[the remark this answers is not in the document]
luke-kiernan: Pretty sure this "revert on failed solve" has been fixed. Resolving this comment and the other similar ones.

rodrigomha:
> Should this change if we have discrete_control for SwitchedShunts:
> 
> https://github.com/Sienna-Platform/PowerFlows.jl/blob/03d240a8ecede1822c37192af5217adf0a890c95/src/common.jl#L128
> 
> It seems that we are adding to the injections directly but we want to handle it differently?

@jd-lara @luke-kiernan @mcllerena @jarry7 We need to verify how is this behavior when the flag for control is enabled or not. We know for sure that when the control flag is enabled is not added to the Ybus in PSS/E. We are not sure what happen when is disabled.

I will review the delta thing if it's respect to the Ybus base + initial condition or simply Ybus base.

src/discrete_control/control_continuation.jl
```diff
@@ -146,29 +198,38 @@ function _control_continuation!(
...
-        reliable ? (dVdp[n_taps + j] = s) : (frozen[n_taps + j] = true)
-    end
+    _probe_device_signs!(set.taps, 0, dVdp, frozen, data, ts, pf; kwargs...)
+    _probe_device_signs!(set.shunts, n_taps, dVdp, frozen, data, ts, pf; kwargs...)
+    _probe_device_signs!(
+        set.facts, n_taps + n_shunts, dVdp, frozen, data, ts, pf; kwargs...)
+    _probe_device_signs!(
+        set.phase_shifters, n_volt, dVdp, frozen, data, ts, pf; kwargs...)
     if any(frozen)
         frozen_names = join(
```
luke-kiernan, line 618: Limit output please, so we don't get hundreds of device names: e.g. switch to `@warn` and specify `maxlog`

src/power_flow_method.jl
```diff
@@ -343,11 +343,14 @@ function _trust_region_step(time_step::Int,
     LinearAlgebra.mul!(stateVector.r_predict, J.Jv, stateVector.Δx_proposed)
     stateVector.r_predict .+= stateVector.r
     predicted_reduction = old_residual_norm - sum(abs2, stateVector.r_predict)
-    # Non-positive predicted reduction (e.g. a non-descent fallback step) would give a
-    # NaN/Inf ρ that stalls the solver; force a rejected-step ρ to shrink the trust region.
+    # The dogleg model reduction is non-negative by construction; a non-positive value
+    # here is floating-point cancellation near convergence (‖r‖²≈0). Force a rejected-step
```
luke-kiernan, line 348: > a non-positive value here is floating-point cancellation near convergence

Hmm not sure I agree. At least for LM, I remember running into non-descent steps at far-from-converged x's

src/power_flow_types.jl
```diff
@@ -21,6 +21,24 @@ current-injection formulation is provided separately. The solver and the
...
 abstract type AbstractACPowerFlow{S <: ACPowerFlowSolverType} <: PowerFlowEvaluationModel end
 
+"""An abstract supertype for the persistent per-solve caches stored in
+`PowerFlowData.solver_cache[]`. Concrete subtypes ([`DCSolverCache`](@ref) for the DC/PTDF path,
+`FastDecoupledCache` for the polar fast-decoupled solver) are type-disjoint, so the slot's type
+discriminates which path populated it — no sentinel tag is needed and a cross-use is a plain
+`MethodError` rather than a silent reuse."""
+abstract type SolverCache end
+
+"""Memoized AC-Jacobian sparse structure, stored in its OWN `PowerFlowData` field (not the shared
```
luke-kiernan, line 31: Yet another linear algebra/matrix structure cache. [sigh] I can live with this, but I'd appreciate having a reference (md file in the `dev` folder of the docs?) that breaks down the use case of each cache struct and highlights the distinctions. Because it's starting to get confusing.

src/solve_dc_power_flow.jl
```diff
@@ -167,10 +192,40 @@ function _run_aba_solve!(
...
+_convert_to_range(::Colon) = Colon()
+
+# PERF: cache ref-to-bus-indices dict.
+"""
+    _shift_angles_to_stored_reference!(data, time_steps = :)
+
+The DC solve computes angles relative to 0 at each subnetwork's ref bus; if the
+subnetwork's ref bus has nonzero angle, shift the angles accordingly.
+"""
+function _shift_angles_to_stored_reference!(
```
luke-kiernan: Initially I was confused because I remember there being a very similar function on `main`. Digging into it, looks like a performance refactor of the DC solve is wrapped up in this PR. This function just changed location. Maybe something to add to `claude.md`: avoid bundling bugfixes and perf reworks into big feature PRs. Instead, fix it on separate branch.

luke-kiernan (commented):
High level comments about the algorithm:

1. Will results be consistent across multiple trials? Might get different results based on the order of the components in `set.taps`, and iirc `get_components` doesn't always return the devices in the same order...
2. Before resorting to `n` individual homotopies ["staircase" route `(0,0) -> (1,0) -> (1, 1)`], we could try a single higher-dimensional one [diagonal, `(0,0) -> (1,1)`]. Would save us a lot of NR trials if it works.
3. Missed opportunity: minimal restoration. When restoring devices from snapped back to continuous, run NR after each device and early-exit as soon as it converges again.

PR #388, opened 2026-06-08
relax linesearches compat

luke-kiernan:
Needed for PSI compatibility. Conflict: `PowerModels` has `NLsolve 4.5.1` which requires `NLSolversBase <= 7.10`. `PowerFlows` has `LineSearches >= 7.6` which requires `NlSolversBase >= 8`.


luke-kiernan:
I'm pretty sure this is failing because the registry hasn't been updated yet, so CI doesn't know v0.23.0 of PNM exists.

jd-lara (approved):

PR #393, opened 2026-06-13
add FNDR algorithm

jd-lara:
This PR adds a version of the FDNR algorithm for power flow that matches best implementations. @luke-kiernan we can test this with the Large EI cases.


src/fast_decoupled_matrices.jl
```diff
@@ -0,0 +1,404 @@
...
+end
+
+"""Accessor for the assembled (unfactored) B′ matrix. Used by tests and diagnostics."""
+get_bp_matrix(fd::FDMatrices) = fd.bp
+
+"""Accessor for the assembled (unfactored) B″ submatrix. Used by tests and diagnostics."""
+get_bpp_matrix(c::FDBppCache) = c.bpp
+
+# -------------------------------------------------------------------------------------------
+# Per-arc parameter recovery
```
luke-kiernan, line 121: I'd vote for adding a new matrix type or two to PNM to support the FNDR, instead of taking this roundabout path of deducing the π model parameters from the existing matrices.

src/fast_decoupled_method.jl
```diff
@@ -0,0 +1,1089 @@
...
+formulations. Factors the formulation Jacobian ONCE at `x0` and reuses the factorization
+across every iteration. Exact residual every iteration ⇒ converges to the same solution as
+NR (linear rate). Shared safeguards (non-divergent backtracking with best-state restore,
+BLOWUP, DVLIM, V≈0 abort) protect against the documented FD failure modes."""
+function _fd_fixed_jacobian_power_flow(
+    pf::AbstractACPowerFlow{FastDecoupledACPowerFlow},
+    data::ACPowerFlowData,
+    time_step::Int64;
+    tol::Float64 = DEFAULT_NR_TOL,
+    maxIterations::Int = DEFAULT_FD_MAX_ITER,
```
luke-kiernan, line 358: Oh boy. So many keyword args. I'd consider making a `FDSettings` struct.

src/fast_decoupled_method.jl
```diff
@@ -0,0 +1,1089 @@
...
+            # DVLIM voltage clamp (+ positivity guard) on the applied step.
+            if !isempty(v_state_idx)
+                v_vals = @view sv.x[v_state_idx]
+                _fd_dvlim_clamp!(sv.Δx_nr, v_state_idx, v_vals, fd_dvlim)
+            end
+
+            # apply step, evaluate exact residual (syncs data: V/θ/P/Q)
+            sv.x .+= sv.Δx_nr
+            residual(sv.x, time_step)
+            ss = sum(abs2, residual.Rv)
```
luke-kiernan: perf nitpick: for large vectors, apparently (Claude code) `dot(v, v)` has an edge because BLAS.

src/PowerFlowData.jl
```diff
@@ -137,16 +137,23 @@ struct PowerFlowData{
...
+    #     object or backend changes — see `_get_or_build_solver_cache!`), `cache` is the actual
+    #     `PFLinearSolverCache`, and `scratch` is the per-solve buffers.
+    #   * AC path, FastDecoupled `:decoupled` solver (`ACPowerFlowData`): a tagged tuple
+    #     `(FD_CACHE_TAG, ::FastDecoupledCache)` holding the factored B′ (once per data/scheme/
+    #     backend) and per-PQ-set factored B″ submatrices (see `_get_or_build_fd_cache!`). The
+    #     `FD_CACHE_TAG` symbol marker keeps the two uses disjoint — the FD getter `error`s loudly
+    #     if it finds a non-FD value (e.g. the DC tuple), so a future cross-use fails fast.
+    # The DC and AC data types never reach each other's getter, so the slot is collision-free
+    # today; the tag is belt-and-suspenders against future drift.
     solver_cache::Base.RefValue{Any}
```
luke-kiernan: The excuse given here

> typed `Any` because the concrete cache types are defined in later

can easily be solved by forward-declaring an abstract supertype, say `SolverCache`. If we wrap the DC tuple in a `DCSolverCache` which subtypes that, and make `FastDecoupledCache <: SolverCache`, then we can write `solver_cache::Ref{Union{Nothing, SolverCache}}`. Still an abstract type, but cleaner: no more sentinel `FD_CACHE_TAG` and no cross-use `error()`

src/fast_decoupled_method.jl
```diff
@@ -0,0 +1,1089 @@
...
+    _validate_fd_settings(pf, fd_variant, fd_scheme, handoff_solver)
+
+Data-free validation of the [`FastDecoupledACPowerFlow`](@ref) settings. Throws a descriptive
+`ArgumentError` on an invalid `fd_variant`/`fd_scheme`, on a `:decoupled` request against a
+non-polar formulation (decoupled is polar-only in v1; rectangular/mixed decoupled is the gated
+WP7), or on an unsupported `handoff_solver`. Returns `nothing` when the configuration is valid.
+"""
+function _validate_fd_settings(
+    pf::AbstractACPowerFlow{FastDecoupledACPowerFlow},
+    fd_variant::Symbol,
```
luke-kiernan: `fd_variant::Symbol` and `fd_scheme::Symbol`: should probably both be scoped enums or types (if we want multiple dispatch).

src/fast_decoupled_method.jl
```diff
@@ -0,0 +1,1089 @@
...
+    FastDecoupledCache
+
+Factor-once cache for the polar :decoupled FD loop, stored in `data.solver_cache[]` as
+`(FD_CACHE_TAG, cache)`. Holds the invalidation key, the constant [`FDMatrices`](@ref) (recovered
+params + factored B′ + assembled B″_full), the `pvpq`-invariant half-step buffers/index vectors
+(factored ONCE per `(data, scheme, backend)` lifetime), and a `Dict` of per-PQ-set
+[`FDPQData`](@ref) keyed on a bus-type signature. `bp_factor_count`/`bpp_factor_count` count B′ and
+B″ factorizations for testability (factor-once verification).
+
+# Fields
```
luke-kiernan, line 657: These first 3 fields that are marked as "invalidation key" seem clunky. Claude suggests bundling them into one struct and using `Ref` instead of `objectid`.

luke-kiernan (changes requested):
First pass. I figure you're probably pretty short on time: should I address these things myself? I'm pretty set on my first comment, migrating some of the FNDR machinery to PNM.

src/fast_decoupled_method.jl
jd-lara: how would that change the number of kwargs?

src/fast_decoupled_matrices.jl
jd-lara: Lets open an issue to do this in a later PR.

src/fast_decoupled_method.jl
jd-lara: A dedicated `FDSettings` wouldn't reduce the number of knobs (the real complaint), and it'd make FD the only solver that doesn't take its settings as flat `solver_settings` kwargs (NR/TR/LM/GD/homotopy all do). The two that deserved to be first-class — variant and scheme — are now type parameters (`FastDecoupledACPowerFlow{V,S}`), which removes them from the kwargs entirely. The rest are advanced safeguards that I'd keep flowing through `solver_settings` like every other solver's, for consistency. Open to revisiting if we ever standardize a settings-struct pattern across all solvers.

jd-lara:
@luke-kiernan this PR needed unsurprisingly better zero impedance management once I tested it with the real raw files. The Argentina case has pretty high zero impedance values in the power flow. After profiling and performance testing, I think this PR is quite ready to go.

src/lcc_utils.jl
luke-kiernan, line 450: Extra negative sign here is a little confusing. Would like to confirm that we have test cases that cover this.

src/lcc_utils.jl
```diff
@@ -507,6 +506,128 @@ function _lcc_jacobian_scalars(
...
+    Vm_tb::Float64,
+)
+    J = zeros(Float64, 4, 4)
+    if iszero(data.lcc.i_dc[i, time_step])
+        # 0-current converter: F = [tap_r - setpoint, tap_i - setpoint, α_r - min, α_i - min].
+        J[1, 1] = 1.0
+        J[2, 2] = 1.0
+        J[3, 3] = 1.0
+        J[4, 4] = 1.0
+        return J
```
luke-kiernan: nitpick: could probably just `return collect(LinearAlgebra.I(4))`.

luke-kiernan (approved):

PR #395, opened 2026-06-19
Clear bus_hvdc_net_power in clear_injection_data!

luke-kiernan:
Zero `bus_hvdc_net_power` each solve, alongside the other injection fields, so construction-time native HVDC setpoints don't double-count and per-solve writes don't accumulate.


jd-lara:
unconditionally zeroing it inside clear_injection_data! introduces a silent regression in the DC PF, so I'm proposing we close this in favor of handling the reset on the PowerSimulations side in https://github.com/Sienna-Platform/PowerSimulations.jl/pull/1637 

I tested the fix on the Argentina system too and corrected the issues found

PR #397, opened 2026-06-22
fix: recover default v35 header to avoid erroring when loading in psse

mcllerena:


jd-lara (approved):

src/psse_export.jl
```diff
@@ -90,6 +90,27 @@ const PSSE_V35_EXTRA_GROUPS = [
 const PSSE_RAW_BUFFER_SIZEHINT = 1024
 const PSSE_MD_BUFFER_SIZEHINT = 1024
 
+# Default PSS/E v35 system-wide data block
+const PSSE_V35_DEFAULT_SYSTEM_WIDE_DATA = """GENERAL, THRSHZ=0.0001, PQBRAK=0.7, BLOWUP=5.0, MaxIsolLvls=4, CAMaxReptSln=20, ChkDupCntLbl=0
```
jd-lara, line 94: @luke-kiernan I think some of thse should actually be consistent with the PowerFlows algorithm used but that is another effort

PR #401, opened 2026-06-24
vsc controls

### LINKED ISSUE #209: Implement VSC Model for AC-HVDC Power Flow model

jd-lara:
This adds the VSC controls for TwoTerminal and also an extension for MT-HVDC with a new object for the MT-HVDC. It might need more validations  

Requires https://github.com/Sienna-Platform/PowerSystems.jl/pull/1704


src/vsc_utils.jl
```diff
@@ -0,0 +1,741 @@
...
+    @inbounds for c in 1:nconv
+        node = dcn.converter_dc_node_ix[c]
+        ix = dcn.converter_ac_bus_ix[c]
+        Vdc = dcn.node_vdc[node, time_step]
+        F[base + node] += _vsc_pdc(dcn, c, Vm[ix], time_step) / Vdc
+    end
+    return
+end
+
+# ── Jacobian derivative helpers (branch on the control mode) ──────────────────────────────────
```
rodrigomha, line 767: I think it would be helpful to have a markdown that summarizes all of these methods for computing the derivatives of things and how they are populated in the infrastructure. 

Overall this looks good, but I can't follow all the math and cases depending on the controllers.

src/mixed_cpb_power_flow_jacobian.jl
```diff
@@ -465,6 +474,108 @@ function _update_mixed_cpb_jacobian_values!(
...
+        _set_entries_for_vsc_mixed!(
+            Jv, Jvnz, diag_base_nz, dcn, e_state, f_state,
+            view(data.bus_type, :, time_step), bus_state_offset, total_bus_state,
+            size(data.lcc.p_set, 1), time_step,
+        )
+    end
+    return
+end
+
+# VSC tail Jacobian (MCPB). Identical to the rectangular writer except the bus current-injection
```
luke-kiernan: "identical except:" refactor to combine?

src/vsc_utils.jl
```diff
@@ -0,0 +1,741 @@
...
+        node = dcn.converter_dc_node_ix[c]
+        mode = dcn.converter_mode[c]
+        P = dcn.p_c[c, time_step]
+        Vdc = dcn.node_vdc[node, time_step]
+        F[2 * c - 1] = _vsc_r1(mode, dcn, c, P, Vdc, time_step)
+        F[2 * c] = dcn.q_c[c, time_step] - dcn.q_set[c, time_step]
+    end
+    base = 2 * nconv
+    G = dcn.G_dc
+    @inbounds for k in 1:nnode
```
luke-kiernan: Note for readability: nested loops compute matmul `I_dc = G * V_dc`.

src/vsc_utils.jl
```diff
@@ -0,0 +1,741 @@
...
+        write_back!(y)
+        _vsc_warm_residual!(F, dcn, Vm, time_step)
+        norm(F) < tol && break
+        _vsc_warm_jacobian!(J, dcn, Vm, time_step)
+        # The warm-start is best-effort: a singular/ill-conditioned tail must never abort the solve,
+        # so on any linear-solve failure we keep the best seed so far and let the joint Newton run.
+        local Δ
+        try
+            Δ = J \ F
+        catch
```
luke-kiernan: Eeek completely generic `catch`. Limit to `SingularException` or similar. Also would be good to have tests.

src/vsc_utils.jl
```diff
@@ -0,0 +1,741 @@
...
+        mode = dcn.converter_mode[c]
+        P = dcn.p_c[c, time_step]
+        Q = dcn.q_c[c, time_step]
+        Vdc = dcn.node_vdc[node, time_step]
+        F[vsc_off + 2 * c - 1] = _vsc_r1(mode, dcn, c, P, Vdc, time_step)
+        F[vsc_off + 2 * c] = _vsc_r2(mode, dcn, c, Q, Vm[ix], time_step)
+    end
+    base = vsc_off + 2 * nconv
+    G = dcn.G_dc
+    @inbounds for k in 1:nnode
```
luke-kiernan: matches earlier unrolled matmul of `I_dc = G * V_dc`. (1) create shared helper? (2) brief 1-line explanatory comment is helpful

src/vsc_utils.jl
```diff
@@ -0,0 +1,741 @@
...
+
+# Derivatives of P_dc w.r.t. (P_c, Q_c, |V_ac|) — nonzero off-P only when the converter has losses.
+function _vsc_pdc_derivatives(dcn::DCNetwork, c::Int, Vm_ac::Float64, time_step::Int)
+    P = dcn.p_c[c, time_step]
+    Q = dcn.q_c[c, time_step]
+    Vmf = sqrt(max(Vm_ac * Vm_ac, V_FLOOR2))
+    S = sqrt(P * P + Q * Q)
+    Ic = S / Vmf
+    dloss_dIc = dcn.loss_b[c] + 2.0 * dcn.loss_c[c] * Ic
+    if iszero(S)
```
luke-kiernan: Might want `abs(S) < eps` instead of `iszero`

src/rectangular_ci_power_flow_jacobian.jl
```diff
@@ -589,6 +652,93 @@ function _update_rect_ci_jacobian_values!(
...
+    @inbounds for b in 1:n_dc_branches(dcn)
+        f = dcn.branch_from[b]
+        t = dcn.branch_to[b]
+        Jv[base + f, base + t] = G[f, t]
+        Jv[base + t, base + f] = G[t, f]
+    end
+    @inbounds for k in 1:nnode
+        Jv[base + k, base + k] = G[k, k]
+    end
+    @inbounds for c in 1:nconv
```
luke-kiernan: nitpick: using complex numbers would cut code verbosity in half. (Avoid complex division for perf reasons, though.) and the math here is long and complicated enough that imo it's worthwhile

src/mixed_cpb_power_flow_jacobian.jl
```diff
@@ -465,6 +474,108 @@ function _update_mixed_cpb_jacobian_values!(
...
+        vk = base + k
+        mode = dcn.converter_mode[c]
+        e = e_state[ix]
+        f = f_state[ix]
+        D = max(e * e + f * f, V_FLOOR2)
+        Vm = sqrt(D)
+        P = dcn.p_c[c, time_step]
+        Q = dcn.q_c[c, time_step]
+        Vdc = dcn.node_vdc[k, time_step]
+        num_r = P * e + Q * f
```
luke-kiernan: another spot where there's enough going on that I'd consider switching to complex numbers for readability

src/vsc_parameters.jl
```diff
@@ -0,0 +1,92 @@
+# VSC converter + DC-network data model.
+#
+# Every PSY DC component (point-to-point `TwoTerminalVSCLine`, and — for multi-terminal grids —
+# `InterconnectingConverter`/`DCBus`/`TModelHVDCLine`) lowers into ONE internal `DCNetwork`. The
+# Newton solver only ever sees `DCNetwork`: converter active/reactive injections `(P_c, Q_c)` and
+# free DC-node voltages `V_dc` are appended as tail unknowns after the bus and LCC blocks, mirroring
+# the LCC tail-unknown pattern. M1 (point-to-point VSC) is the isolated-2-node special case; M2
+# (multi-terminal DC) is the same struct with more nodes/branches and the same kernels.
+
+# Converter control modes. An `Int8`-backed `@enum` stored directly on `DCNetwork`: reading
```
luke-kiernan, line 10: Correct conclusion but confusing phrasing. "dispatches the mode-specific kernels (`_vsc_r1`/`_vsc_r2`/…)" there's no dispatching here, only branching

luke-kiernan (changes requested):
Requests:
1. Test PSS(e) exporter via round-trip on a system with a VSC line [if we don't already]
2. ~Jacobian correctness via asymptotic behavior~ already covered
3. Cooperation with network reductions (including ZIL): are we marking VSC endpoints as irreducible?

For detailed stuff: the generic `catch` here is worth fixing. Rest are fairly minor.

> I think it would be helpful to have a markdown that summarizes all of these methods for computing the derivatives of things and how they are populated in the infrastructure.

Seconded.

src/mixed_cpb_power_flow_jacobian.jl
```diff
@@ -465,6 +474,108 @@ function _update_mixed_cpb_jacobian_values!(
...
+        dIr_df = (Q * D - num_r * 2.0 * f) / D2
+        dIi_de = (-Q * D - num_i * 2.0 * e) / D2
+        dIi_df = (P * D - num_i * 2.0 * f) / D2
+        if bus_types[ix] == PSY.ACBusTypes.PQ
+            # imag-first: F[off] = Ii, F[off+1] = Ir
+            Jvnz[diag_base_nz[1, ix]] += dIi_de
+            Jvnz[diag_base_nz[2, ix]] += dIi_df
+            Jvnz[diag_base_nz[3, ix]] += dIr_de
+            Jvnz[diag_base_nz[4, ix]] += dIr_df
+            Jv[off, pc] = f / D
```
rodrigomha: Another aspect regarding readibility. It could be convenient to also define the methods _vsc_dvr1_de and _vsc_dvr1_df, similar of what we do with the polar. Then vsc_utils.jl can be better organized in sections for polar and rectangular, rather than writing the equations for the mixed and rectangular directly here.

Then improving the docstrings there to highlight that the derivatives are like that is helpful.

rodrigomha:
Similar to my preview comments a markdown like this would be helpful as documentation:

---

## 3. Converter physics (shared across all formulations)

**Sign convention.** $P_c$ is the active power the converter *injects into its AC bus*. It therefore *draws* $P_{dc} = P_c + P_\text{loss}$ from its DC node, and injects DC current $-P_{dc}/V_{dc}$ into the node.

**Converter current magnitude** (with a floor $V_\text{FLOOR2}=10^{-16}$ on $|V_\text{ac}|^2$):

$$
I_c = \frac{\sqrt{P_c^2 + Q_c^2}}{|V_\text{ac}|}, \qquad |V_\text{ac}| = \sqrt{\max(e^2+f^2,\, V_\text{FLOOR2})}
$$

(In polar, $|V_\text{ac}|$ is the bus magnitude directly; in rect/MCPB it is $\sqrt{e^2+f^2}$.)

**Converter loss** (curve coefficients $a,b,c$ from a `LinearCurve` ⇒ $c=0$ or `QuadraticCurve`):

$$
P_\text{loss} = a + b\,I_c + c\,I_c^2
$$

**DC power drawn from the node:**

$$
\boxed{\,P_{dc} = P_c + P_\text{loss} = P_c + a + b I_c + c I_c^2\,}
$$

Code: `_vsc_pdc` (`vsc_utils.jl:496`).

### 3.1 Derivatives of $P_{dc}$

Let $S = \sqrt{P_c^2+Q_c^2}$, $V_\text{mf}=\sqrt{\max(|V_\text{ac}|^2, V_\text{FLOOR2})}$, $I_c = S/V_\text{mf}$, and $\dfrac{\partial P_\text{loss}}{\partial I_c} = b + 2c\,I_c$.
With $\partial I_c/\partial P_c = (P_c/S)/V_\text{mf}$, $\partial I_c/\partial Q_c = (Q_c/S)/V_\text{mf}$, $\partial I_c/\partial |V_\text{ac}| = -I_c/V_\text{mf}$:

$$
\frac{\partial P_{dc}}{\partial P_c} = 1 + (b+2cI_c)\frac{P_c}{S\,V_\text{mf}}, \quad
\frac{\partial P_{dc}}{\partial Q_c} = (b+2cI_c)\frac{Q_c}{S\,V_\text{mf}}, \quad
\frac{\partial P_{dc}}{\partial |V_\text{ac}|} = -(b+2cI_c)\frac{I_c}{V_\text{mf}}
$$

At $S=0$ the $P_c,Q_c$ derivatives of $I_c$ are set to $0$ (avoids division by zero).
Code: `_vsc_pdc_derivatives` (`vsc_utils.jl:626`). Note the off-$P_c$ derivatives vanish for lossless converters ($b=c=0$).

---

## 4. The five control modes

Each converter contributes **two control rows** $r_1, r_2$. The mode also decides whether the converter pins its DC node's voltage (`fixes_dc_voltage`) and whether $r_2$ pins AC voltage (`controls_ac_voltage`).

| Mode | $r_1$ | $r_2$ | Pins $V_{dc}$? | $r_2$ pins $\lvert V_\text{ac}\rvert$? |
|------|-------|-------|:--:|:--:|
| `ControlPQ` | $P_c - P_\text{set}$ | $Q_c - Q_\text{set}$ | no | no |
| `ControlPVac` | $P_c - P_\text{set}$ | $\lvert V_\text{ac}\rvert^2 - V_\text{set}^2$ | no | yes |
| `ControlVdc` | $V_{dc} - V_\text{set}$ | $Q_c - Q_\text{set}$ | yes (DC slack) | no |
| `ControlVdcQ` | $V_{dc} - V_\text{set}$ | $\lvert V_\text{ac}\rvert^2 - V_\text{set}^2$ | yes (DC slack) | yes |
| `ControlPVdcDroop` | $(V_{dc} - V_\text{set}) + k\,P_c$ | $Q_c - Q_\text{set}$ | no (anchors via droop) | no |

The AC-voltage row uses **raw** $|V_\text{ac}|^2 = e^2+f^2$ (floor-free), matching the rectangular PV-pin convention so the residual and its $-2e/-2f$ derivative stay consistent.
Code: `_vsc_r1` (`vsc_utils.jl:505`), `_vsc_r2` (`vsc_utils.jl:516`).

### 4.1 Control-row derivatives

$$
\frac{\partial r_1}{\partial P_c} =
\begin{cases}
1 & \text{PQ, PVac}\\
k & \text{droop}\\
0 & \text{Vdc, VdcQ}
\end{cases}
\qquad
\frac{\partial r_1}{\partial V_{dc}} =
\begin{cases}
1 & \text{Vdc, VdcQ, droop}\\
0 & \text{otherwise}
\end{cases}
$$

$$
\frac{\partial r_2}{\partial Q_c} =
\begin{cases}
0 & \text{AC-voltage modes (PVac, VdcQ)}\\
1 & \text{otherwise}
\end{cases}
\qquad
\frac{\partial r_2}{\partial |V_\text{ac}|} =
\begin{cases}
2\,|V_\text{ac}| & \text{AC-voltage modes}\\
0 & \text{otherwise}
\end{cases}
$$

In rectangular/MCPB the AC-voltage row's derivative is split onto the bus state: $\partial r_2/\partial e = 2e$, $\partial r_2/\partial f = 2f$.
Code: `_vsc_dr1_dP`, `_vsc_dr1_dVdc`, `_vsc_dr2_dQ`, `_vsc_dr2_dVm` (`vsc_utils.jl:592–623`).

---

## 5. DC-network KCL rows

For each DC node $k$, Kirchhoff's current law (current leaving through DC branches = current injected by converters):

$$
\boxed{\ \sum_{j} G_{dc}[k,j]\,V_{dc,j} \ +\  \sum_{c\,\in\,k} \frac{P_{dc,c}}{V_{dc,k}} \ =\  0\ }
$$

The first term is the DC conductance matrix times the DC voltages; the second sums every converter on node $k$.
$G_{dc}$ is the dense nodal conductance built from branch conductances $g_b = 1/r_b$ (a zero-resistance line falls back to $g = 10^6$):

$$
G_{dc}[f,f]\mathrel{+}=g,\quad G_{dc}[t,t]\mathrel{+}=g,\quad G_{dc}[f,t]\mathrel{-}=g,\quad G_{dc}[t,f]\mathrel{-}=g
$$

Code: residual `_set_vsc_tail_residuals!` (`vsc_utils.jl:552`) / `_set_vsc_tail_residuals_rect!` (`:671`); matrix build `_build_G_dc` (`:82`).

### 5.1 DC-KCL row derivatives

$$
\frac{\partial \text{KCL}_k}{\partial V_{dc,k}} = G_{dc}[k,k] - \frac{P_{dc}}{V_{dc,k}^2}, \qquad
\frac{\partial \text{KCL}_k}{\partial V_{dc,j}} = G_{dc}[k,j]\;\;(j\neq k)
$$

$$
\frac{\partial \text{KCL}_k}{\partial P_c} = \frac{1}{V_{dc,k}}\frac{\partial P_{dc}}{\partial P_c}, \qquad
\frac{\partial \text{KCL}_k}{\partial Q_c} = \frac{1}{V_{dc,k}}\frac{\partial P_{dc}}{\partial Q_c}, \qquad
\frac{\partial \text{KCL}_k}{\partial |V_\text{ac}|} = \frac{1}{V_{dc,k}}\frac{\partial P_{dc}}{\partial |V_\text{ac}|}
$$

In rect/MCPB the last term is chain-ruled onto $(e,f)$:

$$
\frac{\partial \text{KCL}k}{\partial e} = \frac{1}{V_{dc,k}}\frac{\partial P_{dc}}{\partial |V_\text{ac}|}\frac{e}{|V_\text{ac}|}
$$

and likewise for $f$:

$$
 \partial|V_\text{ac}|/\partial e = e/|V_\text{ac}|.
$$

Note the $-P_{dc}/V_{dc}^2$ diagonal term is accumulated with `+=` onto $G_{dc}[k,k]$.

src/vsc_utils.jl
```diff
@@ -35,22 +35,57 @@ function _vsc_control_mode(dc_control, ac_control)
...
 
 # Union-find connected components over DC branches; validate each connected DC subnet has at least
 # one slack node (or is wholly droop-controlled). Fails fast naming the offending subnet.
 function _validate_dc_slacks(dcn::DCNetwork)
     nn = n_dc_nodes(dcn)
-    nn == 0 && return
+    iszero(nn) && return
     parent = collect(1:nn)
-    find(i) = (parent[i] == i ? i : (parent[i] = find(parent[i])))
+    function find(i)
```
luke-kiernan, line 88: perf nitpick: iirc this iterative version is often faster in practice
```julia
while parent[i] != i
    parent[i] = parent[parent[i]]
    i = parent[i]
end
return i
```

luke-kiernan (commented):
One minor perf comment. Otherwise looks good. Still need to validate against PSS(e) on a smaller system, though--on the EI, the mismatch at the VSC buses is _worse_ (~300 vs ~3) on this branch than on main, which says something's up.

jd-lara:
### Code review — correctness & performance

Reviewed the full VSC effort for correctness and precompile/compute performance (5 parallel review passes plus `@code_warntype`/`@allocated` verification). **No issues found.** The per-iteration hot path is type-stable and allocation-free by measurement.

**Performance (verified empirically):**
- `DCNetwork` is fully concrete and stored as `dc_network::Base.RefValue{DCNetwork}` (concrete, empty-network placeholder), with every compute site `has_dc_network`-guarded before the loop — no `Union{Nothing}` leaks into inference. [PowerFlowData.jl](https://github.com/Sienna-Platform/PowerFlows.jl/blob/b6ddc4c690ef909088d25baa50206449c3c35086/src/PowerFlowData.jl#L141-L143)
- Rect/mixed VSC Jacobian tail writes exclusively through the prebuilt `VSCJacobianNZCache` nzval indices — zero `Jv[r,c]` setindex in the Newton loop, honoring the `O(N + n_LCC)` contract. [vsc_parameters.jl](https://github.com/Sienna-Platform/PowerFlows.jl/blob/b6ddc4c690ef909088d25baa50206449c3c35086/src/vsc_parameters.jl#L94-L101)
- Per-iteration kernels (`_set_vsc_tail_residuals!`, `_apply_vsc_bus_injections_*!`, `_set_entries_for_vsc_rect_mcpb!`) measured at **0 bytes** allocated after warmup; all allocating work is confined to one-time setup.
- Every kernel returns a consistent `Float64`/`NTuple{4,Float64}` (all literals dotted); enum-value dispatch on `converter_mode::Int8` with no boxing; kernels are concrete top-level functions, no abstract `::Function`-field calls. The `_vsc_pdc`/`_vsc_pdc_derivatives` merge removes the paired double-compute.

**Correctness:**
- Residual↔Jacobian consistency holds for all five control modes; the droop sign `(V_dc − V_set) − k·P_c`, `∂r1/∂P_c = −k` agrees across docs, enum comment, residual, and derivative.
- `V_FLOOR2` flooring is applied identically in residual and Jacobian; polar/rect/mixed share the DC-tail kernels (parity by construction) with a consistent `imag_first_pq` swap; offsets consistent with no off-by-one after the LCC block; loss model and DC-KCL signs exact.

**Conventions:** compliant — no `isa`/`<:` branching, dispatch-based mode/loss handling, no compat bumps.

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

src/post_processing.jl
```diff
@@ -590,6 +590,99 @@ function _compute_segment_flows(
     return entries
 end
 
+# Write the solved VSC / MTDC state back to the PSY components. VSC lines are keyed by
+# reduction-mapped arc tuple, ICs by (AC bus number, DC bus number); parallel components sharing a
```
luke-kiernan, line 594: Edge case: can we have 2 ICs in parallel, such that they share (AC bus number, DC bus number)?

src/post_processing.jl
```diff
@@ -590,6 +590,99 @@ function _compute_segment_flows(
...
+# key are consumed positionally (lowering and write-back iterate the same collections).
+function _write_vsc_solution!(
```
luke-kiernan, line 596: We have 2 reporting options, `Arc` and `Branch`. Doesn't look like `_write_vsc_solution!` differenitates between the two. Should it do so?

src/psse_export.jl
```diff
@@ -2641,21 +2654,21 @@ function _compute_vsc_converter_fields(
...
-    MINLOSS = get_ext_key_or_default(vscline, "MINLOSS_$suffix", psse_converter_loss)
+        REMOT = PSY.get_remote_bus_control_to(vscline)
+        RMPCT = PSY.get_rmpct_to(vscline)
+    end
+
+    # Invert the parser's loss normalization. BLOSS = proportional_term × rated_dc_voltage (kV);
+    # `rated_dc_voltage == 0` (hand-built) keeps the legacy p.u.-power reading. The PSS/E constant
+    # loss splits into ALOSS + MINLOSS, but only the sum is a model quantity (the curve's constant
+    # term), so the whole constant is exported as ALOSS with MINLOSS = 0 (re-parse recovers the
+    # same curve).
```
luke-kiernan, line 2665: Hmm. Worth briefly considering if there's something more suitable here than always taking `MINLOSS = 0`.

luke-kiernan (approved):
3 small comments. imo the branch-reporting vs arc-reporting one is worth fixing. The other 2 are very minor

PR #406, opened 2026-07-02
Candidate fix: switched-shunt control should target PSS/E band edge, not midpoint

luke-kiernan:
## Summary
- Stacked on `jd/discrete-control-homotopy`. Opening this as a separate PR for review rather than folding it into the parent branch, since it changes intentional, tested behavior from the original discrete-control algorithm design.
- **Observed discrepancy**: on a 3-bus repro (switched shunt starting off, `initial_status=[0]`), PSS/E converges to 2 engaged steps (50 Mvar) and bus voltage `0.96077`. `PowerFlows.jl` with `control_discrete_devices=true` converges to 3 steps and `1.0110` — a ~0.05 pu gap.
- **Root cause**: PSS/E's MODSW=1/2 switched-shunt control switches the *minimum number of steps* needed to bring the controlled bus back inside `[VSWLO, VSWHI]`. It does not drive to the band's midpoint. The current implementation sets `vset = (VSWLO+VSWHI)/2` once at construction and treats the shunt as a point-voltage regulator, which can engage more steps than PSS/E would.
- **Change**: `_apply_shunt_deadband!` (new, `control_continuation.jl`), called once per solve right after the base case converges: a shunt already inside `[VSWLO, VSWHI]` is frozen (reusing the existing `frozen` mechanism — PSS/E would not touch it); one outside the band gets its point target set to the nearest violated edge, so the existing sigmoid/relaxation continuation only pushes it back to the edge, not the center. `ControlledSwitchedShunt` gains two new fields (`vset_lo`, `vset_hi`) to carry the band through to solve time.
- Also skips the 2-solve plant-sign probe for shunts already frozen this way (cheap, and consistent with how unreliable-probe freezing already skips stepping).
- The pre-existing "unreliable probe" freeze warning now excludes deadband-held shunts (that's expected/routine, not a fault) so it isn't misleadingly triggered on ordinary in-band shunts.

## Test plan
- [x] New `discrete control: shunt deadband` unit test (band-below/in-band/band-above cases) — 6/6 pass.
- [x] Full existing `test_discrete_control.jl` suite (all shunt/tap/FACTS/PAR testsets, `ramp completes, tight regulation`, `tap+shunt converges (NR)`, etc.) — all green, no regressions (`Main.PowerFlowsTests | 182`, no fail/error column).
- [x] Have not re-validated against the original 3-bus PSS/E repro end-to-end in this session (parser-only + unit-level verification so far) — recommend doing so before merge.

### DIFF
--- a/src/discrete_control/control_continuation.jl
+++ b/src/discrete_control/control_continuation.jl
@@ -156,6 +156,27 @@ function _step_device!(
     return abs(reached - p_now)
 end
 
+# PSS/E's MODSW=1/2 switched-shunt control switches the minimum number of steps needed to
+# bring the controlled bus back inside [VSWLO, VSWHI] — it does not drive to the band's
+# midpoint. Call once per solve, right after the base-case converges and before any device
+# is probed/stepped: freeze a shunt already inside the band, otherwise point `vset` at the
+# nearest violated edge so the continuation below pushes it back to the edge, not the center.
+function _apply_shunt_deadband!(
+    shunts, offset::Int, frozen::Vector{Bool}, data, ts::Int,
+)
+    for (i, d) in enumerate(shunts)
+        y = measured_value(d, data, ts)
+        if y < d.vset_lo
+            d.vset = d.vset_lo
+        elseif y > d.vset_hi
+            d.vset = d.vset_hi
+        else
+            frozen[offset + i] = true
+        end
+    end
+    return nothing
+end
+
 # Probe each device in one concrete vector for its plant sign (dV/dp) and write the result into
 # the shared per-device state at `offset + i`. A function barrier: the body specializes on the
 # concrete element type, so there is no dynamic dispatch over a heterogeneous device collection.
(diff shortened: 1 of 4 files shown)


luke-kiernan:
End-to-end validation against the original 3-bus PSS/E repro (switched shunt starting off, `initial_status=[0]`):

| | PSS/E (solved) | Before this fix | After this fix |
|---|---|---|---|
| Shunt steps engaged | 2 (50 Mvar) | 3 (75 Mvar) | 2 (50 Mvar) |
| Bus 104 Vm | 0.96077 | 1.0110 (Δ +0.0502) | 0.960861 (Δ +0.00009) |

Confirms the deadband fix: the shunt now converges to the same step count PSS/E uses, targeting the violated band edge (`vset_lo=0.95`) rather than the old midpoint (`1.0`), and the solved voltage lands within 0.0001 pu of PSS/E's converged value.

jd-lara (approved):

PR #407, opened 2026-07-07
Reactive power control loops

jd-lara:
This is a clean PR for the effort with the reactive power control


src/discrete_control/controlled_devices.jl
```diff
@@ -0,0 +1,390 @@
...
+continuation. With `PolarNRCache` symbolic reuse this stays O(1) per continuation even as
+`inner_solves` grows; without it, it tracks `inner_solves`. 0 when built without control."""
+get_control_symbolic_factor_count(data) =
+    data.controlled_devices === nothing ? 0 : data.controlled_devices.symbolic_factors[]
+
+# Hot-path instrumentation. The `nothing` (non-control) case — every ordinary AC solve — is a
+# single branch and early return, mirroring `_ctrl_solve!`. Only the continuation path counts.
+@inline function _count_symbolic_factor!(data)
+    cd = data.controlled_devices
+    cd === nothing || (cd.symbolic_factors[] += 1)
```
jd-lara: use `isnothing`

src/discrete_control/controlled_devices.jl
```diff
@@ -0,0 +1,390 @@
...
+    new_tap = p * cis(d.alpha)
+    o = d.nz_offsets
+    @inbounds begin
+        A.nzval[o[1]] += d.yt / abs2(new_tap) - d.yt / abs2(old_tap)
+        A.nzval[o[2]] += -d.yt / conj(new_tap) - (-d.yt / conj(old_tap))
+        A.nzval[o[3]] += -d.yt / new_tap - (-d.yt / old_tap)
+        # Y22 = Yt is tap-independent; no update needed.
+    end
+    d.current = p
+    return nothing
```
jd-lara: no need to do `return nothing` that's outdated julia

src/discrete_control/controlled_devices.jl
```diff
@@ -0,0 +1,390 @@
...
+# row are preserved; `d.synced` tracks the reflected value). Reported flows are computed from
+# these matrices post-loop, so an unsynced branch would report flows at its original tap/angle.
+function _sync_branch_arc_rows!(
+    Yft::SparseArrays.SparseMatrixCSC,
+    Ytf::SparseArrays.SparseMatrixCSC,
+    d::Union{ControlledTap, ControlledPhaseShifter},
+    arc_row::Dict{Tuple{Int, Int}, Int},
+    ix_to_number::Dict{Int, Int},
+)
+    d.current == d.synced && return nothing
```
jd-lara:
```suggestion
    d.current == d.synced && return
```

src/discrete_control/controlled_devices.jl
```diff
@@ -0,0 +1,390 @@
...
+    fb = ix_to_number[d.from_ix]
+    tb = ix_to_number[d.to_ix]
+    ff_new, ft_new, tf_new = _branch_terms(d, d.current)
+    ff_old, ft_old, tf_old = _branch_terms(d, d.synced)
+    r = get(arc_row, (fb, tb), 0)
+    if r != 0
```
jd-lara: This code can be simplified to a call to `haskey`

src/discrete_control/controlled_devices.jl
```diff
@@ -0,0 +1,390 @@
...
+            Yft.nzval[_nz_index(Yft, r, d.from_ix)] += ff_new - ff_old
+            Yft.nzval[_nz_index(Yft, r, d.to_ix)] += ft_new - ft_old
+            Ytf.nzval[_nz_index(Ytf, r, d.from_ix)] += tf_new - tf_old
+            # Ytf(r, to_ix) holds the tt term (= yt): parameter-independent, skipped.
+        end
+    else
+        # The branch may be stored under the reversed arc orientation; the from/to roles
+        # of the four terms swap accordingly (its "from side" is our to bus).
+        r = get(arc_row, (tb, fb), 0)
+        if r == 0
```
jd-lara: same as above

src/linear_solver_backend.jl
```diff
@@ -28,12 +28,15 @@ mutable struct PardisoLinSolveCache
...
 """Union of the KLU, AppleAccelerate, and MKLPardiso solver caches. Every member is concrete so
-the 3-way union stays within Julia's small-union splitting; the KLU member is pinned to
-`J_INDEX_TYPE` because that is the only instantiation flowing through these methods (the AC Newton
-cache and its fallback are built from `J.Jv::SparseMatrixCSC{Float64, J_INDEX_TYPE}`)."""
+the 4-way union stays within Julia's small-union splitting. Both KLU index types are listed: the
+AC Newton cache and its fallback are built from `J.Jv::SparseMatrixCSC{Float64, J_INDEX_TYPE}`
+(`Int32` off Apple, `Int64` on Apple), while PNM's DC ABA factorization is always
+`KLULinSolveCache{Float64, Int64}` regardless of platform — so the DC solve path needs the `Int64`
+member even where `J_INDEX_TYPE === Int32`."""
 const PFLinearSolverCache =
```
jd-lara, line 36: Do not use unions of more than 3. Needs a generalization

jd-lara (commented):
@luke-kiernan This had to incorporate some refactoring caching to avoid multiple refactors per tap outer iteration making the code more complicated

docs/src/explanation/discrete_control.md
```diff
@@ -0,0 +1,304 @@
...
+
+Per-device data problems (unresolvable controlled buses, degenerate tap/step
+ranges, unsupported MODSW modes, implausible voltage setpoints) never abort
+construction: the device is de-enrolled with a `@warn` and stays at its current
+setting — the same *warn and lock* posture PSS/E takes for bad control data.
+
+`ControlledFACTS` and `ControlledPhaseShifter` are fully implemented but their
+metadata sourcing is not yet production-validated (see
+[Metadata sourcing](@ref discrete-control-metadata)); they are enrolled only
+when `solver_settings[:experimental_controls] = true`. A PAR whose
```
jd-lara: This is no longer true after the changes in https://github.com/Sienna-Platform/PowerSystems.jl/pull/1705

docs/src/explanation/discrete_control.md
```diff
@@ -0,0 +1,304 @@
...
+that implicit embedding for the initial implementation. Instead it wraps the
+existing inner solvers in an **outer-loop continuation** (the sigmoid law and
+steepness schedule follow the paper; the outer-loop reinterpretation and its
+convergence safeguards are original to PowerFlows.jl):
+
+  - The inner solver ([`NewtonRaphsonACPowerFlow`](@ref),
+    [`TrustRegionACPowerFlow`](@ref)) is called without any modification.
+  - Between outer iterations only `data` is mutated: Y-bus `nzval` entries for
+    tap/PAR devices; the reactive constant-impedance withdrawal matrix for
+    shunt/FACTS devices.
```
luke-kiernan: Phrasing is a little strange. Something like "only the following fields of `data` are mutated" would be clearer

src/linear_solver_backend.jl
jd-lara: I should target this in a separate PR

docs/src/explanation/discrete_control.md
```diff
@@ -0,0 +1,325 @@
...
+  - `AbstractBranchControl` — devices that mutate the branch's 2×2 Y-bus block:
+    `ControlledTap` (voltage-controlling `TapTransformer`).
+  - `AbstractShuntControl` — devices that mutate the bus reactive
+    constant-impedance withdrawal: `ControlledSwitchedShunt`
+    (voltage-controlling `SwitchedAdmittance`) and `ControlledFACTS`
+    (SVC/STATCOM).
+
+The runtime container is `ControlledDeviceSet`, which holds one concretely
+typed `Vector` per family. All outer-loop traversal iterates each vector
+separately, so dispatch is monomorphic; the per-device kernels
```
luke-kiernan: `monomorphic` is that word actually useful here or just AI flexing its vocab?

docs/src/explanation/discrete_control.md
```diff
@@ -0,0 +1,325 @@
...
+itself allocates small snapshot buffers, which is immaterial next to the inner
+solves).
+
+For `ControlledTap`, `apply_parameter!` rewrites three of the four cached `nzval`
+entries of the sparse Y-bus in place (Y11, Y12, Y21) using cached linear offsets
+(`nz_offsets::NTuple{4,Int}`) resolved once at device-set construction; Y22 = Yt
+is tap-independent and is skipped. The complex tap includes the winding-group
+phase shift `α = PSY.get_α(tx)`, matching PowerNetworkMatrices' stamping
+`t = p·e^{iα}`. The delta update `nzval[k] += Y_new − Y_old` preserves any
+parallel-branch contributions already in the shared slot; `d.current` (not the
```
luke-kiernan: what's `d` here?

docs/src/explanation/discrete_control.md
```diff
@@ -0,0 +1,325 @@
...
+
+## The sigmoid control law
+
+The continuous target for each device is a sigmoid function of the regulated
+quantity ``y`` (the controlled-bus voltage magnitude):
+
+```math
+\sigma(\ell, h, S, y, y_\text{set}) = \frac{h - \ell}{1 + e^{S(y - y_\text{set})}} + \ell
+```
+
```
luke-kiernan, line 108: Is there a way to make it so these values blow in automatically, so that if they change in the code, they change here too?

src/discrete_control/control_continuation.jl
```diff
@@ -0,0 +1,1017 @@
+# Target slope of the local iteration map near the fixed point (0<m<1 ⇒ monotone,
+# non-oscillatory). 0.5 trades settling speed for a 2× margin on the worst-case gain bound.
+# It also bounds the relaxation factor itself: ω = (1−θ)/(1+gbound) ≤ 1−θ = 0.5.
+const CONTROL_CONTRACTION = 0.5
+
+# Snapshot/restore the state a rolled-back trial must not permanently change. Beyond V/θ this
```
luke-kiernan: some of these comments are longer than the body of the function itself, which imo is a red flag

src/discrete_control/control_continuation.jl
```diff
@@ -0,0 +1,1017 @@
...
+function _freeze_device!(frozen::Vector{Bool}, idx::Int, d, ts::Int, reason::String)
+    frozen[idx] = true
+    @warn "discrete control: freezing device $(d.name) at parameter \
+        $(current_parameter(d)) (time step $ts): $reason"
+    return
+end
+
+# Refresh one ControlledFACTS's voltage-dependent susceptance bound (`_facts_b_limit`) from
+# its measured controlled-bus voltage, once per continuation pass, BEFORE this pass's damped
+# target/clamp reads `parameter_limits`. Chatter guard: `frozen` (this device's CURRENT
```
luke-kiernan: comment verbosity: "chatter guard"

src/discrete_control/control_continuation.jl
```diff
@@ -0,0 +1,1017 @@
...
+        numeric_refactor!(ctx.cache, ctx.J.Jv)
+    catch e
+        e isa LinearAlgebra.SingularException || rethrow()
+        return false
+    end
+    return true
+end
+
+# ∂F/∂p into `rhs` (zeroed first). Returns `true` iff the family has an analytic polar form here;
+# `false` signals the caller to use the FD probe (no family currently returns `false`, but the
```
luke-kiernan: > no family currently returns `false`

still accurate?

src/discrete_control/control_continuation.jl
```diff
@@ -0,0 +1,1017 @@
...
+# only (state layout x[2b−1]=Vm, x[2b]=Va and ∂F/∂p are polar-specific); rect/mixed formulations
+# fall back to the FD `_plant_sign`. Signs are validated against the FD probe in the tests.
+
+# Sensitivity context: residual+Jacobian built at the CURRENT converged base state and numerically
+# factored (reusing P1's persisted symbolic factorization). Built ONCE per continuation (probe
+# phase) and thereafter kept current by `_refresh_sensitivity_context!` (values-only, no rebuild)
+# after each batched-pass joint solve. `nothing` ⇒ the linear path is unavailable (non-polar
+# formulation, or the base Jacobian is singular) and the caller uses FD probes.
+struct _SensitivityContext{C, R, JT}
+    cache::C
```
luke-kiernan: We have enough different varieties of "cache" objects in PF that I'd consider renaming to something more descriptive, like `lin_cache`.

Also, what's "P1" here?
> reusing P1's persisted symbolic factorization

src/discrete_control/control_continuation.jl
```diff
@@ -0,0 +1,1017 @@
...
+    # on every subsequent batched pass and does NOT count here — the Jacobian structure (and thus
+    # the persisted symbolic factorization this counter tracks) is topology-invariant across the
+    # continuation, so only this one build should ever register.
+    _count_symbolic_factor!(data)
+    n = length(residual.Rv)
+    return _SensitivityContext(
+        cache, residual, J, zeros(n), zeros(n), copy(view(data.bus_type, :, ts)))
+end
+
+# Values-only refresh at a new converged base state: the Jacobian STRUCTURE depends only on
```
luke-kiernan: comment verbosity: "jacobian structure" gets 3 mentions in these 10 lines.

src/discrete_control/control_continuation.jl
```diff
@@ -0,0 +1,1017 @@
...
+#
+# `P_net`/`Q_net`/`P_net_set` and the four constant-I/Z withdrawal vectors are all captured from
+# `data` ONCE at `ACPowerFlowResidual` construction and are NOT independently re-read by the
+# residual functor: `_update_residual_values!`'s PQ-bus case does `P_net[ix] += ...` (a
+# TELESCOPING correction from the bus's voltage at the residual's LAST evaluation to its new
+# value), which is only correct when the SAME residual object is the one driving every
+# evaluation of a single NR solve. Here `ctx.residual` is evaluated once per PASS while `data`'s
+# voltage is moved BETWEEN passes by the joint solve's OWN, unrelated residual object — so
+# `P_net`/`Q_net` must be rebuilt from `data` (exactly like the constructor) before every
+# evaluation, or they silently accumulate a wrong, ever-growing correction. Likewise
```
luke-kiernan, line 195: Might be worth opening an issue to revisit. The current machinery was built with ZIP loads in mind, not FACTS devices or switched shunts.

src/discrete_control/control_continuation.jl
```diff
@@ -0,0 +1,1017 @@
...
+    _fill_shared!(n_shared, n_taps + n_shunts, counts, set.facts)
+
+    # Single reused snapshot buffer for the probe phase, the pass loop, and the post-loop
+    # per-device restore (`_plant_sign`, `_continuation_to!`, `_batched_pass!`, `_restore_one!`):
+    # those phases run one after another and are never simultaneously live, so one buffer
+    # suffices. `snap_and_restore!`'s own `pre` snapshot is the exception — it stays live WHILE
+    # it calls `_restore_one!`, so it gets its own separate buffer, never this one.
+    scratch_snap = _snapshot_state(data, ts)
+
+    # Build the linearized-sensitivity context ONCE (one numeric factorization reusing P1's
```
luke-kiernan: What's P1?

src/discrete_control/control_metadata.jl
```diff
@@ -0,0 +1,361 @@
...
+        name = PSY.get_name(tx)
+        arc = PSY.get_arc(tx)
+        fb = PSY.get_number(PSY.get_from(arc))
+        tb = PSY.get_number(PSY.get_to(arc))
+        md = _tap_metadata(tx, tb)
+        fix = _resolve_bus_ix(bus_lookup, reverse_bus_search_map, fb)
+        tix = _resolve_bus_ix(bus_lookup, reverse_bus_search_map, tb)
+        cix = _resolve_bus_ix(bus_lookup, reverse_bus_search_map, md.cbus)
+        if isnothing(fix) || isnothing(tix)
+            missing_bus = tb
```
luke-kiernan, line 165: nitpick: use case for `something`

src/discrete_control/controlled_devices.jl
```diff
@@ -0,0 +1,395 @@
...
+controlled_bus_ix(d::ControlledFACTS) = d.controlled_ix
+
+# Result columns meaningful only for FACTS; taps/shunts report the neutral value so the
+# frame stays rectangular (dispatch, not an `isa` branch, over the mixed device loop).
+result_delivered_q_mvar(::AbstractControlledDevice, data, ts::Int) = missing
+result_delivered_q_mvar(d::ControlledFACTS, data, ts::Int) =
+    delivered_q_mvar(d, data.bus_magnitude[d.bus_ix, ts])
+result_saturated(::AbstractControlledDevice) = false
+result_saturated(d::ControlledFACTS) = d.saturated
+voltage_setpoint(d::ControlledTap) = d.vset
```
luke-kiernan: If it's the same on all subtypes, just define it on the supertype.

src/ac_power_flow_jacobian.jl
```diff
@@ -107,19 +105,57 @@ J(time_step)  # Updates the Jacobian matrix stored internally in J.
 J.Jv  # Access the Jacobian matrix stored internally in J.
 ```
 """
+# Memoize the expensive Jacobian sparse-structure build (~3.2 MB on 2000 buses) in the
```
luke-kiernan: super verbose comment

src/solve_ac_power_flow.jl
```diff
@@ -19,6 +19,12 @@ The bus types can be changed from PV to PQ if the reactive power limits are viol
 ## Keyword Arguments
 - `tol`: Infinite norm of residuals under which convergence is declared. Default is `1e-9`.
 - `maxIterations`: Maximum number of Newton-Raphson iterations. Default is `30`.
+- `write_device_settings`: Write the solved tap ratios / shunt admittances / phase-shifter
```
luke-kiernan: > when the solve ran with `control_discrete_devices`, write-back happens regardless of this flag

Wait, what's the use case then? By my reading:
- `control_discrete_devices = false` means "don't adjust discrete devices." Writing back to the system is a no-op, so the setting is redundant.
- `control_discrete_devices = true` already writes back to the system, so the setting is also redundant.

src/solve_ac_power_flow.jl
```diff
@@ -65,6 +78,77 @@ function solve_and_store_power_flow!(
...
+Write the solved discrete-control device settings back into the `PSY.System`:
+tap ratios (`set_tap!`), switched-shunt admittances (`set_Y!`/`set_initial_status!`,
+convention-aware — see below), and phase-shifter angles (`set_α!`). FACTS devices
+carry no stored setting field in PSY and are skipped. A device no longer present
+in `system` is skipped with a `@warn` (its solved setting is not written back).
+Mutates the user's system — called by [`solve_and_store_power_flow!`](@ref)
+whenever the solve ran with active discrete controls (a self-consistent system
+requires it), and when `write_device_settings = true` otherwise.
+
+Switched shunts are sourced under one of two conventions (see
```
luke-kiernan: Is this explained in another spot? I feel like it is. In which case should just link to that other spot.

src/solve_ac_power_flow.jl
```diff
@@ -256,10 +340,29 @@ function _ac_power_flow(
         end
     end
 
-    @error(
-        "could not enforce reactive power limits after $MAX_REACTIVE_POWER_ITERATIONS"
+    # Iteration cap reached: the last `_check_q_limit_bounds!` flipped one or more PV buses to
```
luke-kiernan, line 335: Just want to confirm that this (hit max q iterations => return converged = true and warn, instead of converged = false and error) is a purposeful design decision, not AI going rogue.

luke-kiernan (approved):
Superficial stuff mostly.

src/discrete_control/control_continuation.jl
jd-lara: Done

src/solve_ac_power_flow.jl
jd-lara: This is on purpose to tighten up main's behavior here. Hitting the Q-limit cap logged an @error but still returned converged=true, with the state left inconsistent. This makes the behavior safe.

PR #410, opened 2026-07-09
Fix DC power-flow slack balance

### LINKED ISSUE #404: Standalone DCPowerFlow power-balance update
Issue:
* Slack is not adjusting at the defined slack-bus for a standalone DCPowerFlow
* To establish a physically meaningful solution - a simple implementation could be to adjust the slack at the REF bus(es) so the power balance is balanced again.
* A better solution is to apply distributed slack adjustment.

Example:
* Using the RTS_GMLC in PSCB, the power balance is not balanced.  Specifically, 8550 MW of load while generation is 6006 MW.  

Image

jd-lara:
Closes #404 and I will make the updates to support the distributed slack in this branch

### DIFF
--- a/src/solve_dc_power_flow.jl
+++ b/src/solve_dc_power_flow.jl
@@ -107,6 +107,62 @@ function _shift_angles_to_stored_reference!(
 end
 
 """
+    _distribute_dc_slack!(data)
+
+Distribute each subnetwork's active-power imbalance across its participating
+buses (including the reference bus) in proportion to `bus_slack_participation_factors`,
+writing the result into `bus_active_power_injections`. No-op unless the user configured
+distributed slack (explicit factors or headroom mode), detected via a
+non-empty `computed_generator_slack_participation_factors`.
+"""
+function _distribute_dc_slack!(
+    data::Union{PTDFPowerFlowData, vPTDFPowerFlowData, ABAPowerFlowData},
+)
+    isempty(get_computed_gspf(data)) && return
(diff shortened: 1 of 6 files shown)


src/solve_dc_power_flow.jl
```diff
@@ -366,6 +474,29 @@ function solve_power_flow(
...
+Solve the DC power flow for `pf` and store the solution back into `system`:
+bus angles, unit voltage magnitudes, and redistributed active generation at
+the slack/participating buses. Returns `true` (DC always converges). Only the
+first time step is written back.
+"""
+function solve_and_store_power_flow!(
+    pf::AbstractDCPowerFlow,
+    system::PSY.System;
+    linear_solver::Union{Nothing, AbstractString} = nothing,
+    max_iterations::Int = DEFAULT_NR_MAX_ITER,
```
luke-kiernan, line 489: why do we have `max_iterations` here?

edit: oh, it's "max iterations for power redistribution." Let's give the kwarg a better name, or at least make its default value something like `DEFAULT_REDIST_MAX_ITER`.  (Using `DEFAULT_NR_MAX_ITER` here is just plain confusing...)

luke-kiernan (approved):
One confusing naming choice. Rest looks good

src/solve_dc_power_flow.jl
jd-lara: I'll fix it on the rebase in the reactive power control PR

PR #413, opened 2026-07-13
psi integration

jd-lara:
This PR provides support for multi-period AC PF with the new reactive power components. Needed for https://github.com/Sienna-Platform/PowerSimulations.jl/pull/1648

@luke-kiernan it also incorporates a fix to the LCC modeling indentified when using this feature with large scale texas planning cases

### DIFF
--- a/docs/src/explanation/discrete_control.md
+++ b/docs/src/explanation/discrete_control.md
@@ -51,11 +51,11 @@ with an `ArgumentError` for combinations that are not yet supported:
     inner solvers. (FastDecoupled factors B′/B″ once and would silently reuse
     them after a tap move; LM/GD/Homotopy are unvalidated as continuation inner
     solvers.)
-  - **Single time step:** `time_steps > 1` is rejected — device state does not
-    yet track per-time-step baselines (taps mutate the shared Y-bus; shunt
-    deltas are per-column).
   - **No LCC HVDC:** the continuation's rollback does not yet cover the
     per-time-step LCC state.
+
+`time_steps > 1` is supported for every device family (see
+[Multiperiod solves](@ref discrete-control-multiperiod)).
 
 Per-device data problems (unresolvable controlled buses, degenerate tap/step
 ranges, unsupported MODSW modes, implausible voltage setpoints) never abort
@@ -64,6 +64,30 @@ setting — the same *warn and lock* posture PSS/E takes for bad control data.
 
 `ControlledFACTS` is enrolled unconditionally (no flag); its metadata sourcing
 caveats are noted in [Metadata sourcing](@ref discrete-control-metadata).
+
+## [Multiperiod solves](@id discrete-control-multiperiod)
+
+Each time step regulates independently. `ControlledDeviceSet` carries a
+per-time-step store (one column per step, sized at construction); around each
+step's continuation the solver loads that step's device state into the scalar
+scratch the engine mutates and saves it back afterwards. Shunt/FACTS state is a
+plain scalar swap — their control deltas land in the per-column withdrawal
+matrices. Taps mutate the *shared* Y-bus, so each step instead resets the tap to
+its enrollment baseline via a Y-bus delta-update before regulating
+(reset-to-baseline design), and branch flows are computed inside the time-step
+loop so every step's flows reflect that step's tap position.
+
+Two multiperiod-specific behaviors:
+
+  - **No PSY write-back:** a PSY component holds a single scalar setting, so
+    [`write_device_settings!`](@ref) skips (with a warning) for
+    `time_steps > 1`. Per-step settings are reported by
+    [`get_controlled_device_results`](@ref), one row per device per time step
+    (`time_step` column).
+  - **Same-snapshot seeding:** time-invariant inputs (constant-power,
+    constant-current, and constant-impedance withdrawals, injections, reactive
+    limits) are seeded into every column from the system snapshot; callers with
+    time-varying data overwrite columns before solving.
 
 ## Device abstraction
 
@@ -322,7 +346,8 @@ pinned at `b_lim` while its regulated bus remains off `voltage_setpoint` is
 flagged `saturated = true` (with a warning). `get_controlled_device_results`
 carries that flag plus `delivered_q_mvar = b·|V|²·S_base` on the FACTS rows, and
 under active controls the solved `Q` is written back to the component's
-`reactive_power_required`.
+`reactive_power_required` (single time step only; see
+[Multiperiod solves](@ref discrete-control-multiperiod)).
 
 ## Reserved seams for future work
 
--- a/src/ac_power_flow_jacobian.jl
+++ b/src/ac_power_flow_jacobian.jl
@@ -783,8 +783,10 @@ function _set_entries_for_lcc(data::ACPowerFlowData,
             _calculate_dQ_dα_lcc(s.tap_r, s.i_dc, xtr_r, Vm_fb, phi_r, alpha_r) # ∂Q_fb/∂α_fb
         Jv[idx_p_tb, idx_tap_to] = dP_dt_tb # ∂P_tb/∂t_tb
         Jv[idx_p_tb, idx_angle_to] = s.dP_dα_tb # ∂P_tb/∂α_tb
+        # Inverter dQ: −xtr_i flips the commutation-chain term to the inverter sign
+        # (see _lcc_jacobian_scalars); the leading sin ϕ_i term is x_t-free.
         Jv[idx_q_tb, idx_tap_to] =
-            _calculate_dQ_dt_lcc(s.tap_i, s.i_dc, xtr_i, Vm_tb, phi_i) # ∂Q_tb/∂t_tb
+            _calculate_dQ_dt_lcc(s.tap_i, s.i_dc, -xtr_i, Vm_tb, phi_i) # ∂Q_tb/∂t_tb
         # φ_i convention flips sign of ∂φ_i/∂α_i vs the rectifier; negate helper output.
         Jv[idx_q_tb, idx_angle_to] =
             -_calculate_dQ_dα_lcc(s.tap_i, s.i_dc, xtr_i, Vm_tb, phi_i, alpha_i) # ∂Q_tb/∂α_tb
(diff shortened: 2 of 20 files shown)


src/lcc_utils.jl
```diff
@@ -10,7 +10,15 @@ function _calculate_ϕ_lcc(
     x_t::Float64,
     Vm::Float64,
 )::Float64
-    raw = sign(I_dc) * (cos(α) - (x_t * I_dc) / (sqrt(2) * Vm * t))
+    # Commutation-drop term uses abs(I_dc): the drop x_t·|I_dc| always REDUCES the
```
luke-kiernan: Super verbose comment. Same for the 2 following ones. These belong in commit message or perhaps docs (take a look at the "lcc derivatives markdown")

luke-kiernan (approved):
Didn't look too closely at the tests or run them locally, but the PR straightforward enough that I don't feel the need. The one thing I'd suggest: go trim some of AI's comments

m-bossart (approved):
Agree with Luke about the comments; could be made more concise. There are also a lot of comments that describe the LCC bug fix, instead of just describing the corrected code. These types of comments make less sense moving forward in the codebase outside of the context of this PR. 

One other small fix. In `solving_a_power_flow.jl` tutorial, there is a comment that should be removed about not supporting multi-period AC power flows:
https://github.com/Sienna-Platform/PowerFlows.jl/blob/3b113d10414ab0bc4dfbaf4e92e7e8ca1135400c/docs/src/tutorials/solving_a_power_flow.jl#L77

docs/src/explanation/discrete_control.md
```diff
@@ -65,6 +65,30 @@ setting — the same *warn and lock* posture PSS/E takes for bad control data.
...
+
+Each time step regulates independently. `ControlledDeviceSet` carries a
+per-time-step store (one column per step, sized at construction); around each
+step's continuation the solver loads that step's device state into the scalar
+scratch the engine mutates and saves it back afterwards. Shunt/FACTS state is a
+plain scalar swap — their control deltas land in the per-column withdrawal
+matrices. Taps mutate the *shared* Y-bus, so each step instead resets the tap to
+its enrollment baseline via a Y-bus delta-update before regulating
+(reset-to-baseline design), and branch flows are computed inside the time-step
+loop so every step's flows reflect that step's tap position.
```
luke-kiernan, line 78: Give this paragraph a read and an edit. (Phrasing of "around each...saves it back afterwards;" flowery vocab of "delta-update before regulating.")

src/discrete_control/control_continuation.jl
```diff
@@ -965,16 +965,21 @@ end
...
     get_controlled_device_results(data) -> DataFrames.DataFrame
 
-Solved discrete-control device settings: one row per enrolled device with its family,
-name, control band, enrollment-time (`initial`) and solved (`final`) parameter. The
-solved settings are written back to the `PSY.System` by
-[`solve_and_store_power_flow!`](@ref) under active controls, and applied to PSS/E
-exports by [`update_exporter!`](@ref). Returns an empty frame when the data was built
-without discrete control.
+Solved discrete-control device settings: one row per enrolled device per time step, with
+its family, name, time step, control band, enrollment-time (`initial`) and solved
```
luke-kiernan, line 969: > enrollment-time (`initial`) and solved (`final`) parameter for that step.

My first thought: wait, what's enrollment time? [They mean "enrollment-time _parameter_," implied via omitted comma.] Just say "initial parameter"

src/discrete_control/control_continuation.jl
```diff
@@ -983,27 +988,31 @@ function get_controlled_device_results(data)
...
-                push!(saturated, result_saturated(d))
+        for ts in 1:get_time_steps(data)
+            for (fam, devices) in (
+                ("TapTransformer", set.taps),
+                ("SwitchedAdmittance", set.shunts),
+                ("FACTSControlDevice", set.facts),
+            )
+                for (i, d) in enumerate(devices)
+                    lo, hi = stored_parameter_limits(set, d, i, ts)
+                    push!(family, fam)
```
luke-kiernan, line 999: nitpick: could switch to something more self-documenting, like a vector of named tuples or a wrapper around a tuple of vectors [better column access]. But meh it's fine.

src/discrete_control/controlled_devices.jl
```diff
@@ -74,6 +74,16 @@ struct ControlledDeviceSet
     taps::Vector{ControlledTap}
     shunts::Vector{ControlledSwitchedShunt}
     facts::Vector{ControlledFACTS}
+    # Per-time-step scalar state, indexed [device, ts], each column seeded from the freshly-built
+    # device's fields; `load_device_state!`/`save_device_state!` swap them in/out of the device
+    # scratch `_control_continuation!` mutates. Taps mutate the SHARED Y-bus, so their store is
```
luke-kiernan: Searching for "shared Y-bus" gives 6 matches. Deduplicate comments

src/discrete_control/controlled_devices.jl
```diff
@@ -128,13 +218,40 @@ end
 
 controlled_bus_ix(d::AbstractControlledDevice) = d.controlled_ix
 
+# Results reporting reads the PERSISTED per-ts store, not the live device scalar (which
```
luke-kiernan, line 220: verbose comment. Also, what the comment doesn't say that imo would be helpful: "current" in its electrical meaning (_not_ meaning up-to-date).

src/ac_power_flow_jacobian.jl
```diff
@@ -783,8 +783,10 @@ function _set_entries_for_lcc(data::ACPowerFlowData,
             _calculate_dQ_dα_lcc(s.tap_r, s.i_dc, xtr_r, Vm_fb, phi_r, alpha_r) # ∂Q_fb/∂α_fb
         Jv[idx_p_tb, idx_tap_to] = dP_dt_tb # ∂P_tb/∂t_tb
         Jv[idx_p_tb, idx_angle_to] = s.dP_dα_tb # ∂P_tb/∂α_tb
+        # Inverter dQ: −xtr_i flips the commutation-chain term to the inverter sign
+        # (see _lcc_jacobian_scalars); the leading sin ϕ_i term is x_t-free.
         Jv[idx_q_tb, idx_tap_to] =
-            _calculate_dQ_dt_lcc(s.tap_i, s.i_dc, xtr_i, Vm_tb, phi_i) # ∂Q_tb/∂t_tb
+            _calculate_dQ_dt_lcc(s.tap_i, s.i_dc, -xtr_i, Vm_tb, phi_i) # ∂Q_tb/∂t_tb
```
luke-kiernan, line 789: Do you know why our current jacobian tests failed to catch this error? [And did/could you add a test case to guard against regression...]

src/initialize_power_flow_data.jl
```diff
@@ -47,8 +47,11 @@ function initialize_power_flow_data!(
         removed_buses,
         sys,
     )
-    data.bus_active_power_injections[:, 1] .= bus_active_power_injections
-    data.bus_reactive_power_injections[:, 1] .= bus_reactive_power_injections
+    # Seed EVERY column from the snapshot (a column vector broadcasts across time): each step
```
luke-kiernan: Deduplicate/simplify "seed every column" comments--like "apply baseline to all columns" would be sufficient

src/post_processing.jl
```diff
@@ -1242,6 +1244,48 @@ function lcc_results_dataframe(
...
+"""
+    get_hvdc_results(sys::PSY.System, data::ACPowerFlowData)
+
+Per-time-step HVDC results from a solved AC power flow, as a NamedTuple of DataFrames:
+`lcc` ([`PowerSystems.TwoTerminalLCCLine`](@extref) lines), `vsc` (point-to-point
+[`PowerSystems.TwoTerminalVSCLine`](@extref) lines), `mtdc_converters`
+([`PowerSystems.InterconnectingConverter`](@extref)), and `mtdc_lines`
+([`PowerSystems.TModelHVDCLine`](@extref) DC branches). Each table is the corresponding
+single-time-step results frame with a `time_step` column prepended, concatenated over
+`1:get_time_steps(data)`; a family absent from the system yields an empty table. Powers are in
```
luke-kiernan, line 1266: > Each table is the corresponding single-time-step results frame with a `time_step` column prepended, concatenated over `1:get_time_steps(data)`

For our overall results, we return a dictionary, `t => (NamedTuple of DataFrames)`, where each `DataFrame` has axes `(device name, [powers/voltages/etc])`. This looks like it's formatted differently, "long" results instead of "wide"...

edit: confirmed. A little inconsistent, but I can live with it. imo would marginally help to mention "long results style" in the docstring

src/power_flow_types.jl
```diff
@@ -87,14 +87,15 @@ function _reject_fd_decoupled_on_nonpolar(
     return
 end
 
-# Discrete device control is validated only for NR/TR inner solvers and a single time step (FD
-# reuses stale B′/B″ after tap moves; LM/GD/Homotopy are unvalidated; device state has no per-ts
-# baseline yet). Centralized so the three formulation constructors cannot drift. NR/TR types are
-# defined later in this file — references resolve at call time.
+# Discrete device control is validated only for NR/TR inner solvers (FD reuses stale B′/B″
```
luke-kiernan, line 90: comment bloat

src/rectangular_ci_power_flow_jacobian.jl
```diff
@@ -899,10 +899,12 @@ function _set_entries_for_lcc_rect!(
         cos_phi_i = cos(phi_i)
         sin_phi_i = sin(phi_i)
         # ∂ϕ derivatives with sin(ϕ)→0 clamp guard (return 0 at clamp).
+        # Inverter uses −xtr_i: the corrected ϕ_i subtracts the commutation drop, flipping
```
luke-kiernan, line 902: How many places do we have a comment like this...reduce to "see `_lcc_utils` for explanation of negative sign" or the like

test/test_discrete_control.jl
```diff
@@ -772,9 +788,10 @@ end
         control_discrete_devices = true)
     @test_throws ArgumentError ACMixedPowerFlow{LevenbergMarquardtACPowerFlow}(;
         control_discrete_devices = true)
-    # Device state does not track per-time-step baselines yet.
-    @test_throws ArgumentError ACPolarPowerFlow(;
-        control_discrete_devices = true, time_steps = 2)
+    # Construction is permissive: every device family supports time_steps>1 (taps via
+    # reset-to-baseline), so a multi-step solver builds without inspecting the device mix.
```
luke-kiernan, line 792: Deduplicate comments: "reset-to-baseline"

luke-kiernan (commented):
Almost entirely superficial stuff. The non-superficial stuff: why did our jacobian tests miss that negative sign?

Also, we should mention in some front-and-center place that "multi-period AC power flow" does _not_ mean that we're reading time series data.

edit: oh, didn't realize I reviewed this already. oops.

src/ac_power_flow_jacobian.jl
jd-lara: Probably because the value test is zero

PR #414, opened 2026-07-13
More non-convergence diagnostics

luke-kiernan:
Some WIP stuff to better diagnose non-convergence


luke-kiernan:
I opened this for visibility, but it's really a bespoke branch for NAERM stuff. With 8 other PRs open I'll close. Can integrate these features later if desired

PR #415, opened 2026-07-15
area interchange and multi-ref 

jd-lara:
This PR enables the use of multiple ref buses in the PowerFlows and the area interchange object. After merging this we can solve all the ERCOT large scale planning cases, testing with MMWG and WECC are pending. 

This PR lands on top of the PSI integration PR


docs/src/explanation/area_interchange.md
```diff
@@ -0,0 +1,131 @@
...
+Solving drives every ``r_a`` to the same tolerance as the bus mismatch rows, so the schedules hold
+exactly at convergence while ``\Delta P_a`` reports how much the area's slack generation moved to
+achieve them.
+
+### Ties
+
+**AC ties** are in-service AC branches whose endpoints lie in different areas (a
+`ThreeWindingTransformer` decomposes into its star-node windings, so a winding crossing the
+boundary is a correctly metered tie). The tie-flow kernel reads each corridor's admittances
+directly from the aggregate Y-bus. Because a nodal Y-bus diagonal sums *every* device at a bus, a
```
luke-kiernan, line 88: Hmmm another spot where we recover the branch values from the ybus...fine for now but worth noting for an eventual refactor

docs/src/explanation/area_interchange.md
```diff
@@ -0,0 +1,131 @@
...
+per-tie `diag_pollution` correction, cached at enrollment, recovers the corridor's own
+self-admittance; this stays exact under a controlled tap on the corridor itself.
+
+**DC ties** are two-terminal HVDC links (LCC or VSC) whose converter buses lie in different
+areas. The metered-terminal converter active power enters ``NI_a`` with the same out-of-area sign
+convention as AC ties. Metering follows the branch's `ext["metered_end"]` (`"from"`/`"to"`,
```
luke-kiernan: Which side is metered for AC? Does that also follow `ext["metered_end"]`?

docs/src/explanation/area_interchange.md
```diff
@@ -0,0 +1,131 @@
...
+
+## Formulation
+
+Each enrolled area contributes one state and one residual row, appended as a tail after the LCC/VSC
+tails:
+
+  - **State** ``\Delta P_a``: an active-power adjustment injected at the area's slack bus (the
+    area-slack machine, PSS/E ISW). It couples into that bus's P-balance row at the same seam as
+    the distributed-slack term.
+  - **Residual** ``r_a = NI_a - PDES_a``, where ``NI_a`` is the sum of metered active power over the
```
luke-kiernan, line 65: I'm not familiar with all this PSS(E) official terminology: NI, PDE, ISW. Worth fixing, or will our target audience be in-the-know?

edit: okay they are defined, but you kind of have to hunt for the definitions. especially for `PDES`

> scheduled target ``PDES_a``
> ...
> the area-slack machine, PSS/E ISW
> ...
> ``NI_a`` is the sum of metered active power over the area's boundary ties

Also I found the use of the term "seam" here odd. Technical term? If not I'd maybe just say "in a similar way to..."
> at the same seam as the distributed-slack term.

docs/src/explanation/area_interchange.md
```diff
@@ -0,0 +1,131 @@
+# Area Interchange Control
+
+## What it does
+
+**Net-interchange control**: each participating `PSY.Area` is driven so that its net
+active-power exchange with the rest of the system meets a scheduled target ``PDES_a`` (derived from
+the system's `PSY.AreaInterchange` records). Control is *embedded* in the AC solver — the
```
luke-kiernan, line 7: What happens if the ISW bus hits its Q limit and gets PV -> PQ switched mid-solve?

docs/src/explanation/area_interchange.md
```diff
@@ -0,0 +1,131 @@
...
+per area enrolled at construction:
+
+| Column            | Meaning                                                                       |
+|:----------------- |:----------------------------------------------------------------------------- |
+| `area`            | PSY area name.                                                                |
+| `ni_solved`       | Achieved net interchange (recomputed from the tie kernels, AC and DC ties).   |
+| `pdes`            | Scheduled target.                                                             |
+| `delta_p`         | Solved ``\Delta P_a`` (`0.0` for a relaxed area).                             |
+| `schedule_status` | `:enforced` or `:relaxed`.                                                    |
+| `beyond_limits`   | Whether ``\Delta P_a`` exceeds the slack bus's machine active-power headroom. |
```
luke-kiernan: If the slack bus machine has nonzero distributed slack weight, that could compound: perhaps ``\Delta P_a`` alone doesn't make it exceed its limits, but ``\Delta P_a`` _plus_ distributed slack does...is that taken into account in `beyond_limits`? [If it's correct in the code then I'm not too concerned that it's omitted here.]

luke-kiernan (commented):
First pass for now, tackling conceptual big stuff. I haven't thought about area interchanges before, so some of the questions may be silly

Also, separately: as we've expanded the amount of "additional rows/columns" for our jacobian, it'd be nice to have a depiction of (1) the overall block structure and (2) the arrangement of stuff within the blocks. See e.g. `src/ac_power_flow_jacobian.jl:420` and `src/ac_power_flow_jacobian.jl:300`.

PR #417, opened 2026-07-24
Add precompilation workload to reduce first-call latency

jd-lara:
## Summary
Adds a precompilation workload that exercises the core AC and DC power-flow solve paths at package build time, moving first-call latency from user sessions into the cached package image.

## Key Changes
- **New `src/precompile.jl`**: Defines a minimal in-memory test system (4 buses, mixed component types covering REF/PV/PQ partitions) and a `PrecompileTools.@compile_workload` block that exercises:
  - AC polar power flow via explicit `PowerFlowData` construction + in-place solve (twice, to compile cache refresh/reuse paths)
  - AC polar power flow via one-shot public API with DataFrame results
  - DC power flow (ABA direct solve)
  - PTDF-based DC power flow
  - All solver logging is silenced to keep precompile output clean
  
- **Exposed `_precompilation_workload_system()`**: Plain function so the test suite can exercise the exact workload inputs and validate precompilation coverage

- **Updated `src/PowerFlows.jl`**: Added `import PrecompileTools` and included `precompile.jl` last (after all solve paths are defined)

- **Updated `Project.toml`**: Added `PrecompileTools` dependency (compat `1.2`)

- **Updated `docs/src/reference/api/internal_solvers.md`**: Added documentation section for the precompilation workload

## Implementation Details
The workload is designed to:
- Avoid I/O and parsing costs by hand-building the system from PSY constructors (same pattern as test utilities)
- Prevent KLU factorization handles from leaking into the package image by keeping all objects local to the workload block
- Be skippable during development via PrecompileTools preference for fast rebuild iterations
- Cover the PSI-stable surface (explicit `PowerFlowData` + in-place solve, plus one-shot API)

https://claude.ai/code/session_01EXkkoqjT6CejCV2AbnhTiP


luke-kiernan:
Too many other tasks on my plate for now, but I hope to come back to this in a week or so

PR #420, opened 2026-07-27
fix: resolve icd get_number_table call for multiple sa associations

### LINKED ISSUE #418: PSSE transformer export assumes first supplemental attribute has get_table_number
## Summary

`PowerFlows.PSSEExporter` appears to assume that the first supplemental attribute on a `TwoWindingTransformer` is an impedance correction attribute with a `PowerSystems.get_table_number` method. This can fail when another supplemental attribute type is attached to a transformer before or instead of `ImpedanceCorrectionData`.

In our case, a production-cost-modeling workflow attaches `PowerSystems.FixedForcedOutage` supplemental attributes to branches, including some `TapTransformer` components, to represent contingency-derived flow gates. When PSSE RAW export runs, the transformer writer tries to call `get_table_number(::FixedForcedOutage)`, causing a `MethodError`.

## Observed error

```julia
MethodError: no method matching get_table_number(::FixedForcedOutage)
The function `get_table_number` exists, but no method is defined for this combination of argument types.

Closest candidates are:
  get_table_number(::ImpedanceCorrectionData)
   @ PowerSystems ...\PowerSystems\vhoCS\src\impedance_correction.jl:75

Stacktrace:
  [1] _write_2w_transformer_record3_winding1!(io::IOBuffer, exporter::PSSEExporter, transformer::TapTransformer)
    @ PowerFlows ...\PowerFlows\4Bw8F\src\psse_export.jl:1022
  [2] write_to_buffers!(exporter::PSSEExporter, ::Val{Symbol("Transformer Data")})
    @ PowerFlows ...\PowerFlows\4Bw8F\src\psse_export.jl:2345
  [3] write_export(exporter::PSSEExporter, name::String; overwrite::Bool)
    @ PowerFlows ...\PowerFlows\4Bw8F\src\psse_export.jl:3178
  [4] solve_power_flow!
    @ PowerFlows ...\PowerFlows\4Bw8F\src\psse_export.jl:3288
  [5] solve_power_flow!(pf_e_data::PowerSimulations.PowerFlowEvaluationData{PSSEExporter}, container::PowerSimulations.OptimizationContainer, sys::System)
    @ PowerSimulations ...\PowerSimulations\Ss5io\src\network_models\power_flow_evaluation.jl:1295
```

## Suspected source

The failing code path appears to be in `src/psse_export.jl`, inside `_write_2w_transformer_record3_winding1!`:

```julia
supp_attr = PSY.get_supplemental_attributes(transformer)
TAB1 = !isempty(supp_attr) ? PSY.get_table_number(supp_attr[1]) : 0
```

This assumes `supp_attr[1]` is an object that supports `PSY.get_table_number`, which seems to be true for `ImpedanceCorrectionData` but not for other valid transformer supplemental attributes.

## Why this matters

`PowerSystems.jl` supplemental attributes are also used by other workflows. For example, `FixedForcedOutage` can be attached to branches to represent contingency/outage metadata. If a transformer has a `FixedForcedOutage` supplemental attribute, PSSE export fails even though the attribute is unrelated to impedance correction table export.

This also makes the exporter dependent on supplemental attribute ordering. If a transformer has both `ImpedanceCorrectionData` and another supplemental attribute, `supp_attr[1]` may not be the impedance correction attribute.

## Expected behavior

PSSE transformer export should only use supplemental attributes that actually represent impedance correction data / table-number metadata. If no such attribute exists, `TAB1` should default to `0`.

A defensive pattern could be something like:

```julia
impedance_correction = findfirst(attr -> attr isa PSY.ImpedanceCorrectionData, supp_attr)
TAB1 = impedance_correction === nothing ? 0 : PSY.get_table_number(supp_attr[impedance_correction])
```

or an equivalent trait/filter that does not depend on supplemental attribute order and does not call `get_table_number` on arbitrary supplemental attributes.

## Minimal reproduction shape

1. Create or load a `PowerSystems.System` with a `TapTransformer` / `TwoWindingTransformer`.
2. Attach a non-impedance supplemental attribute to that transformer, such as `FixedForcedOutage`.
3. Run `PowerFlows.PSSEExporter` / PSSE RAW export.
4. Export fails with `MethodError: no method matching get_table_number(::FixedForcedOutage)`.

## Environment

```julia
(pcm_sim) pkg> st
  [336ed68f] CSV v0.10.16
  [13f3f980] CairoMakie v0.15.13
  [a93c6f00] DataFrames v1.8.2
  [864edb3b] DataStructures v0.19.6
  [87dc4568] HiGHS v1.24.1
  [fc1677e0] HydroPowerSimulations v0.18.0
  [2cd47ed4] InfrastructureSystems v3.6.0
  [b6b21f68] Ipopt v1.15.0
  [94fada2c] PowerFlows v0.24.0
  [bed98974] PowerNetworkMatrices v0.24.2
  [e690365d] PowerSimulations v0.37.1
  [f00506e0] PowerSystemCaseBuilder v2.5.0
  [bcd98974] PowerSystems v5.12.0
⌃ [295af30f] Revise v3.16.1
  [10745b16] Statistics v1.11.1
⌅ [9e3dc215] TimeSeries v0.25.2
  [9e70acf3] Xpress v0.18.1
  [ade2ca70] Dates v1.11.0
  [8ba89e20] Distributed v1.11.0
  [56ddb016] Logging v1.11.0
```

mcllerena:


luke-kiernan (approved):
Do we have a regression test for the bug that prompted this? If so, go ahead and merge; if not, please add one.

PR #423, opened 2026-08-03
multi-ref and area interchange to psy6

m-bossart:
Port of https://github.com/Sienna-Platform/PowerFlows.jl/pull/415
@jd-lara Should the github workflow changes from the PR also be applied to psy6?


jd-lara (approved):

m-bossart:
@jd-lara  The test failures are due to needing this PR which implements the node-breaker parsing but also fixes the RDC scaling issue:  https://github.com/Sienna-Platform/PowerFlowFileParser.jl/pull/28/

PR #426, opened 2026-08-04
Enable LCC + discrete control with multiple timesteps

### LINKED ISSUE #422: Support systems with LCC +  control_discrete_devices=true

m-bossart:
This closes #422

### DIFF
--- a/src/discrete_control/control_continuation.jl
+++ b/src/discrete_control/control_continuation.jl
@@ -4,9 +4,11 @@
 const CONTROL_CONTRACTION = 0.5
 
 # State a rolled-back trial must not permanently change: V/θ, `bus_type` (one-way PV→PQ flips), the
-# injection columns (Q clamps, distributed-slack setpoint updates), and the VSC DC-network state
+# injection columns (Q clamps, distributed-slack setpoint updates), the VSC DC-network state
 # (`_read_vsc_state!` writes the iterate into dcn.p_c/q_c/node_vdc every residual eval, so a
-# diverged trial corrupts it exactly like bus state).
+# diverged trial corrupts it exactly like bus state), and the LCC solver state (converter taps,
+# thyristor angles, DC current). Derived caches (phi, branch_admittances) are re-derived from the
+# restored state rather than snapshotted.
 @inline function _dc_state_cols(data, ts::Int)
     dcn = get_dc_network(data)
     if has_dc_network(dcn)
@@ -37,7 +39,21 @@ end
     )
 end
 @inline function _capture_state!(
-    (vmag, vang, btype, pinj, qinj, dc_p, dc_q, dc_v, lcc_rt, lcc_it, lcc_ra, lcc_ia, lcc_idc),
+    (
+        vmag,
+        vang,
+        btype,
+        pinj,
+        qinj,
+        dc_p,
+        dc_q,
+        dc_v,
+        lcc_rt,
+        lcc_it,
+        lcc_ra,
+        lcc_ia,
+        lcc_idc,
+    ),
     data,
     ts::Int,
 )
@@ -65,7 +81,21 @@ end
 @inline function _restore_state!(
     data,
     ts::Int,
-    (vmag, vang, btype, pinj, qinj, dc_p, dc_q, dc_v, lcc_rt, lcc_it, lcc_ra, lcc_ia, lcc_idc),
+    (
+        vmag,
+        vang,
+        btype,
+        pinj,
+        qinj,
+        dc_p,
+        dc_q,
+        dc_v,
+        lcc_rt,
+        lcc_it,
+        lcc_ra,
+        lcc_ia,
+        lcc_idc,
+    ),
 )
     data.bus_magnitude[:, ts] .= vmag
     data.bus_angles[:, ts] .= vang
--- a/src/discrete_control/control_continuation.jl
+++ b/src/discrete_control/control_continuation.jl
@@ -3,14 +3,9 @@
 # It also bounds the relaxation factor itself: ω = (1−θ)/(1+gbound) ≤ 1−θ = 0.5.
 const CONTROL_CONTRACTION = 0.5
 
-# Solver state a rolled-back trial must not permanently change: V/θ, `bus_type` (one-way
-# PV→PQ flips), the injection columns, and the VSC/LCC tail state. Derived caches (`phi`,
-# `branch_admittances`) are recomputed from the restored state, so they cannot drift from it.
-# An absent VSC or LCC subsystem is carried as empty vectors.
-#
-# Named fields rather than a positional tuple: nearly every column is `Vector{Float64}`, so a
-# mis-ordered or dropped column would be type-invisible and would silently stop being rolled
-# back.
+# State a rolled-back trial must restore. `bus_type` is here because the inner solve's Q-limit
+# logic makes one-way PV→PQ flips. An absent VSC or LCC subsystem is carried as empty vectors,
+# which is what the capture/restore guards test.
 struct ControlStateSnapshot
     vmag::Vector{Float64}
     vang::Vector{Float64}
--- a/src/discrete_control/control_continuation.jl
+++ b/src/discrete_control/control_continuation.jl
@@ -3,12 +3,10 @@
 # It also bounds the relaxation factor itself: ω = (1−θ)/(1+gbound) ≤ 1−θ = 0.5.
 const CONTROL_CONTRACTION = 0.5
 
-# State a rolled-back trial must not permanently change: V/θ, `bus_type` (one-way PV→PQ flips), the
-# injection columns (Q clamps, distributed-slack setpoint updates), the VSC DC-network state
-# (`_read_vsc_state!` writes the iterate into dcn.p_c/q_c/node_vdc every residual eval, so a
-# diverged trial corrupts it exactly like bus state), and the LCC solver state (converter taps,
-# thyristor angles, DC current). Derived caches (phi, branch_admittances) are re-derived from the
-# restored state rather than snapshotted.
+# State a rolled-back trial must not permanently change: V/θ, `bus_type` (one-way PV→PQ
+# flips), the injection columns, and the VSC/LCC tail state, all of which the inner solve
+# rewrites on every residual eval. Derived caches (phi, branch_admittances) are recomputed
+# from the restored state instead, so they cannot drift from it.
 @inline function _dc_state_cols(data, ts::Int)
     dcn = get_dc_network(data)
     if has_dc_network(dcn)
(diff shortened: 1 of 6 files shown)


src/lcc_utils.jl
```diff
@@ -859,16 +859,16 @@ function initialize_LCCParameters!(
...
+    lcc_inverter_tap .= PSY.get_inverter_tap_setting.(lccs)
     # Fixed tap targets used to pin the tap state for 0-current (0-MW) converters.
     data.lcc.rectifier.tap_setpoint .= PSY.get_rectifier_tap_setting.(lccs)
     data.lcc.inverter.tap_setpoint .= PSY.get_inverter_tap_setting.(lccs)
     lcc_dc_line_resistance .=
         PSY.get_r.(lccs) .+ PSY.get_rectifier_rc.(lccs) .+ PSY.get_inverter_rc.(lccs)
     lcc_i_dc .= _lcc_i_dc_from_p_set.(lcc_dc_line_resistance, lcc_p_set)
-    lcc_rectifier_delay_angle[:, 1] .= PSY.get_rectifier_delay_angle.(lccs)
-    lcc_inverter_extinction_angle[:, 1] .= PSY.get_inverter_extinction_angle.(lccs)
+    lcc_rectifier_delay_angle .= PSY.get_rectifier_delay_angle.(lccs)
```
m-bossart, line 870: Is this actually a pre-existing bug fix or was there a reason this was only applied to the first time step that I am missing?

luke-kiernan: Good catch, yep that's a bug. When we added support for multi-period AC power flow (from within PF) the buses, voltages, etc. got updated to broadcast the system value to all time steps, but evidently we missed the LCCs.

src/discrete_control/control_continuation.jl
```diff
@@ -3,30 +3,55 @@
...
         data.bus_active_power_injections[:, ts],
         data.bus_reactive_power_injections[:, ts],
         dc_p, dc_q, dc_v,
+        lcc_rt, lcc_it, lcc_ra, lcc_ia, lcc_idc,
     )
 end
 @inline function _capture_state!(
-    (vmag, vang, btype, pinj, qinj, dc_p, dc_q, dc_v),
+    (
+        vmag,
```
luke-kiernan: If this same tuple order is the same in several places, I'd suggest switching to an immutable struct or named tuple. More self-documenting. Otherwise it's hard to tell which entry to grab for a specific thing: the LCC-associated values start at the...9th entry?

test/test_lcc_discrete_control.jl
```diff
@@ -0,0 +1,362 @@
...
+# `i_dc` is CHECKPOINTED because `initialize_LCCParameters!` derives it from `p_set` at
+# construction and no solve path rewrites it.
+
+const LCC_CHECKPOINTED_FIELDS = Dict(
+    PF.LCCConverterParameters => [:tap, :thyristor_angle],
+    PF.LCCParameters => [:i_dc],
+)
+const LCC_DERIVED_FIELDS = Dict(
+    PF.LCCConverterParameters => [:phi],
+    PF.LCCParameters => Symbol[],  # branch_admittances is a Vector (not per-ts), re-derived
```
luke-kiernan: > branch_admittances is a Vector (not per-ts), re-derived

I'm confused. For that comment, I'd expect to see `Symbol[:branch_admittances]` or the like.

test/test_lcc_discrete_control.jl
```diff
@@ -0,0 +1,362 @@
...
+    va_ctrl = copy(data.bus_angles[:, 1])
+    lcc_taps = copy(data.lcc.rectifier.tap[:, 1])
+
+    # Twin system: same devices fixed at the converged settings, control off. The shunt locks
+    # through its own `Y`, as `write_device_settings!` saves back. The FACTS device needs a
+    # `FixedAdmittance` instead: `contributes_reactive_power(::FACTSControlDevice)` is false,
+    # so `reactive_power_required` reaches reporting only, and the device's physical footprint
+    # is a constant-Z withdrawal that exists solely inside the continuation. Locking via
+    # `set_reactive_power_required!` alone leaves the twin's voltage off by the device's whole
+    # contribution (~5e-4 pu here).
```
luke-kiernan: Very confusing comment. Please trim.

luke-kiernan (commented):
The implementation itself looks fine. However, I'm struggling to make anything of the tests, to tell if they're quality or just token coverage. A good part of this is probably AI's awful commenting practices: it's happy to write verbose paragraph but neglects to provide basic signposts.

src/discrete_control/control_continuation.jl
```diff
@@ -3,61 +3,102 @@
...
-# State a rolled-back trial must not permanently change: V/θ, `bus_type` (one-way PV→PQ flips), the
-# injection columns (Q clamps, distributed-slack setpoint updates), and the VSC DC-network state
-# (`_read_vsc_state!` writes the iterate into dcn.p_c/q_c/node_vdc every residual eval, so a
-# diverged trial corrupts it exactly like bus state).
+# Solver state a rolled-back trial must not permanently change: V/θ, `bus_type` (one-way
+# PV→PQ flips), the injection columns, and the VSC/LCC tail state. Derived caches (`phi`,
+# `branch_admittances`) are recomputed from the restored state, so they cannot drift from it.
+# An absent VSC or LCC subsystem is carried as empty vectors.
+#
+# Named fields rather than a positional tuple: nearly every column is `Vector{Float64}`, so a
```
luke-kiernan: AI comment that I'd remove. "How this used to work" won't be useful or relevant 100 commits down the road

test/test_lcc_discrete_control.jl
```diff
@@ -0,0 +1,381 @@
...
+    sys = build_lcc_control_system()
+    pf = ACPolarPowerFlow(; control_discrete_devices = true, time_steps = time_steps)
+    data = PowerFlowData(pf, sys)
+    _set_lcc_bus101_loads!(data, time_steps)
+    solve_power_flow!(data)
+    @test all(data.converged)
+    # Bus 101 is stiff enough that the per-step loads move voltage magnitude by only ~1e-6,
+    # so bus angle is the quantity that actually separates the steps. Assert that separation
+    # exceeds the parity tolerance, or the comparisons below could pass for any `ts`.
+    @test maximum(abs.(data.bus_angles[:, 1] .- data.bus_angles[:, 3])) > LCC_PARITY_ATOL
```
luke-kiernan: A little odd that you only check separation in angle here. Shouldn't you do the same for bus magnitude too?

test/test_lcc_discrete_control.jl
```diff
@@ -0,0 +1,381 @@
...
+        @test PowerFlows.get_lcc_count(data) > 0
+    end
+end
+
+# ── 2. Single-period control ──────────────────────────────────────────────────────────────
+
+# Twin-parity tolerance, sized between the two quantities it has to separate: the measured
+# twin error is ~1.5e-8 in bus angle, while the per-step spread the multiperiod tests rely on
+# is ~2.4e-5 in bus angle and ~1e-6 in voltage magnitude.
+const LCC_PARITY_ATOL = 1e-7
```
luke-kiernan, line 108: I'm still not a huge fact of the fact that the voltage magnitude spread is only 1 order of magnitude above the tolerance. I had Claude checks the numbers against pencil-and-paper estimates and it found no issues, though, so it's not a big concern.

Aside: odd detail from Claude's math. Not sure if it's relevant, though.
> Note R/X ≈ 5–23 — this network is resistance‑dominated, so the usual P↔θ / Q↔V decoupling is inverted: ΔQ moves angle, and only weakly moves magnitude.

test/test_lcc_discrete_control.jl
```diff
@@ -0,0 +1,381 @@
...
+        pf = ACPolarPowerFlow(; control_discrete_devices = true, time_steps = nts)
+        data = PowerFlowData(pf, sys)   # must not throw
+        @test PowerFlows.get_lcc_count(data) > 0
+    end
+end
+
+# ── 2. Single-period control ──────────────────────────────────────────────────────────────
+
+# Twin-parity tolerance, sized between the two quantities it has to separate: the measured
+# twin error is ~1.5e-8 in bus angle, while the per-step spread the multiperiod tests rely on
```
luke-kiernan: Interesting detail from Claude:

> With DEFAULT_NR_TOL = 1e-9 and Jacobian entries of order 1/x = 1e4 here, converged angle noise should be ~1e‑13. So the 1.5e‑8 is a genuine small modeling difference between the controlled solve and the locked twin — most likely the FACTS constant‑Z footprint reproducing its in‑continuation behavior only approximately.

Is that cause for concern, or expected? i.e. should the in-continuation behavior be reproduced exactly, or are there known, purposeful modeling differences?

luke-kiernan: Okay checks out. Looks like it's a `Float64` vs `Float32` thing: the twin has the extra on the ybus diagonal, which is lower-precision than the FACTS fields. Details if you're interested:
<details>
<summary> Claude's reasoning </summary>

> Both give the bus an injection of b·|V|². The difference is purely where the number is stored. 
> 
> Bus 101's Ybus diagonal from the two incident branches:
> 
> Y₁₁ = 1/(2.3e-3 + j1e-4) + 1/(5.4e-4 + j1e-4) + jb_charging
>     = 2224.41 − j350.38
> 
> The FACTS rails at its STATCOM limit b_lim = min(rating/V, q_cap/V²) (controlled_devices.jl:406‑413) ≈ 9.985 pu. The twin stores imag = −350.38 + 9.985 = −340.40 in Float32; the controlled run stores −350.38 in Float32 and keeps 9.985 exactly in Float64. Both operands sit in the same binade, so the discrepancy is 0 or 1 ulp:
> 
> ulp(Float32) at |B| ≈ 350  =  3.05e-5     ← net error in effective susceptance
> 
> Push that through the same sensitivities as before (ΔV = Z·conj(ΔS/V) = (R+jX)(−jΔQ) ⇒ Δ|V| ≈ X·ΔQ, Δθ ≈ −R·ΔQ; δQ = δB·V², angle channel R ∈ [4.4e‑4, 2.3e‑3]):
> 
> δθ = R·δB  =  1.3e-8 … 7.0e-8 rad
> δVm = X·δB =  2.1e-9 pu
> 
> The measured twin error of 1.5e‑8 in angle sits at the bottom of that band, and the magnitude channel is ~1e‑9 — which is why the comment quotes the twin error only in angle. Both match.
> 
> Two falsifiable signatures if you want to confirm rather than accept the argument:
> 
> 1. The error should be lumpy, not smooth — exactly 0 or 1 ulp depending on where b_lim falls. I checked: b = 10.0 (exactly representable, same binade) gives δB = 0 identically; b = 9.985 gives a full ulp. So nudging max_shunt_current slightly should make the twin error jump between ~0 and ~5e‑8 rather than drift.
> 2. One decisive line, no solve needed: compare imag(data2.power_network_matrix.data[101,101]) - facts_b against imag(data.power_network_matrix.data[101,101]). They should differ by exactly 0 or 3.05e‑5.
</details>
</summary>

luke-kiernan (approved):
A few small things

luke-kiernan (approved):

PR #432, opened 2026-08-19
Port onto the infrastore line; identify the exported system by its UUID

jd-lara:
`IS.get_uuid` no longer exists. The one use was `_validate_same_system`, comparing the two Systems' internals — and neither obvious replacement works, each failing silently in the opposite direction. `IS.get_id` on a System's internal always returns 0, because Systems are never attached to a SystemData and the id defaults to UNASSIGNED_ID, so the guard would accept any two unrelated systems. `PSY.get_system_uuid(exporter.system)` always differs, because `fast_deepcopy_system` forwards name and description but not the uuid, so the exporter's copy carries a fresh one — the old code worked precisely because the internal uuid survives a deepcopy.

So `PSSEExporter` captures the base system's UUID at construction and compares against that. It is never reassigned: validation already guarantees equality, so a fixed value cannot go stale.

Pins move to the infrastore-line branches across all three environments. PowerNetworkMatrices stays pinned to its branch rather than the registry because registry v0.24 requires PowerSystems 5.11, which this line does not satisfy.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -26,19 +26,19 @@ Pardiso = "46dd5b70-b6fb-5a00-ae2d-e8fea33afaf2"
 [extensions]
 PowerFlowsPardisoExt = "Pardiso"
 
-# Pin the unreleased Sienna stack to the psy6/IS4 development branches so the package
+# Pin the unreleased Sienna stack to the feat/infrastore-integration development branches so the package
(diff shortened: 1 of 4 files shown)


luke-kiernan (approved):
looks simple enough as to the changes here in PF

PR #433, opened 2026-08-19
fold detection via sign(det(J)) and bordered matrix

luke-kiernan:
For fold detection, I believe `sign(det(J))` here is the thing to look at, over the sign of the real part of the smallest eigenvalue. This PR achieves that via a bordered matrix strategy, described below by AI:

> Border J with fixed vectors:
> 
> M = [ J   b; cᵀ  d ]
> 
> Schur complement gives det(M) = det(J) · (d − cᵀJ⁻¹b). Define
> 
> s = d − cᵀ J⁻¹ b;  g = 1/s = det(J) / det(M)
> 
> g is smooth in the continuation parameter, and vanishes exactly when J is singular — because det M is a fixed smooth function that is nonzero near a simple fold for generic b, c. So sign(g) differs from sign(det J) only by the constant factor sign(det M), and flips of g are flips of det J.
> 
> Computational cost: one solve J y = b against the existing factorization plus a dot product. 
> 
> g can change sign two ways, and they're distinguishable by magnitude:
> 
> - Through zero (|g| decreases toward 0, then grows): genuine singularity of J. This is your fold.
> - Through a pole (|g| blows up): det M crossed zero — your bordering degenerated, not a property of J. Re-pick b, c and recompute the monitor over the whole interval.

Some light polishing to do--AI comment verbosity, error if feature is requested for non-TR/NR solvers--but I want input on the algorithm

### DIFF
--- a/src/residual_condition_diagnostics.jl
+++ b/src/residual_condition_diagnostics.jl
@@ -174,21 +174,256 @@ end
 # Fold / voltage-collapse bail-out state and the shared per-iteration hook.
 # ---------------------------------------------------------------------------
 
-# UNSEEN = pre-first-observation. A non-finite/non-converged λ_min is deliberately
-# not a sign here; the bail decision treats it as a conservative abort.
-IS.@scoped_enum(EigvalSign, UNSEEN = 0, NEGATIVE = -1, POSITIVE = 1)
+# FIXME a bit over-engineered.
+"""Deterministic pseudo-random unit vector, filled by an LCG so the bordering — and
+therefore every logged monitor value — is reproducible across runs, platforms, and
+Julia versions (no `Random` dependency, whose streams are not version-stable)."""
+function _fill_border_vector!(v::Vector{Float64}, seed::UInt64)
+    s = seed * 0x9E3779B97F4A7C15 + 0x1
+    @inbounds for i in eachindex(v)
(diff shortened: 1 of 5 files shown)


src/residual_condition_diagnostics.jl
```diff
@@ -174,21 +174,256 @@ end
 # Fold / voltage-collapse bail-out state and the shared per-iteration hook.
 # ---------------------------------------------------------------------------
 
-# UNSEEN = pre-first-observation. A non-finite/non-converged λ_min is deliberately
-# not a sign here; the bail decision treats it as a conservative abort.
-IS.@scoped_enum(EigvalSign, UNSEEN = 0, NEGATIVE = -1, POSITIVE = 1)
+# FIXME a bit over-engineered.
```
jd-lara, line 177: Fix before merge

jd-lara (commented):
if we are doing power flow in the loop, how can we report back that the set point and power flow are not solvable?

luke-kiernan:
> if we are doing power flow in the loop, how can we report back that the set point and power flow are not solvable?

I'd be inclined to merge this and address that in a follow-up PR.

jd-lara:
> > if we are doing power flow in the loop, how can we report back that the set point and power flow are not solvable?
> 
> I'd be inclined to merge this and address that in a follow-up PR.
yeah, I'll merge after finishing the rebase of psy6 branch

PR #435, opened 2026-08-19
Make update_state!/update_data! respect ZIP withdrawals

### LINKED ISSUE #431: `update_state!` and `update_data!` ignore ZIP loads
The helpers `update_state!` and `update_data!` haven't been updated in so long that they neglect constant-impedance and constant-current withdrawals. Anything that reconstructs a residual from `data` using those functions in a system with ZIP loads (or `SwitchedAdmittance` components) will give incorrect values.

We should bring the functions in `state_indexing_helpers.jl` up-to-date and/or add some guards for the cases they aren't equipped to handle. Worth adding some test cases too.

luke-kiernan:
Closes #431 for the polar formulation.

## Problem

`update_state!` and `update_data!` built the REF/PV power slots from the bare constant-power field `bus_*_power_withdrawals`, ignoring the constant-current (`∝ |V|`) and constant-impedance (`∝ |V|²`) ZIP terms. Anything reconstructing a residual from `data` in a system with ZIP loads or `SwitchedAdmittance` components got the wrong net injection at REF and PV buses.

This is not confined to diagnostics: `calculate_x0` calls `update_state!`, so it perturbs the starting point of every polar solve on a ZIP system, and `_previous_solution_start` inherits it for the time-step warm start.

## Fix

Route both through `get_bus_active_power_total_withdrawals` / `get_bus_reactive_power_total_withdrawals`, matching how `ACPowerFlowResidual` already forms `P_net`/`Q_net`. Docstrings now state the net-injection convention and that the two functions are inverses.

No ordering hazard in `update_data!`: a bus's ZIP withdrawal depends only on its own magnitude, and REF/PV magnitudes are fixed setpoints, so PQ magnitude writes never race a ZIP read.

## Tests

A 3-bus REF/PV/PQ fixture carrying a full ZIP load at every bus, with off-nominal REF/PV setpoints (1.02, 1.03) so the `|V|` and `|V|²` terms are separable from constant power. Four testsets, including guards that assert the fixture actually has nonzero const-I/const-Z and off-nominal setpoints — without those the tests could not detect the bug.

Verified they discriminate, by stashing the source fix and re-running:

| testset | against unfixed code |
|---|---|
| `update_state!` includes ZIP withdrawals | **3 fail** — REF P `-0.1` vs `-0.616`, REF Q `-0.05` vs `-0.308`, PV Q `-0.05` vs `-0.312` |
| `update_data!` includes ZIP withdrawals | **3 fail** — `0.35` vs `0.866`, etc. |
| `update_state!` at the solution is a residual zero | **1 fail** — `‖Rv‖∞ = 0.516`, exactly the dropped withdrawal |
| round trip | passes either way |

The round-trip testset is deliberately non-discriminating — the bug is symmetric, so it cannot catch it. Kept as an inverse invariant guarding against a future one-sided edit; the other three are what pin the issue.

Full suite green: 46033 passing, 0 failures.

## Scope

Polar only. The rectangular and mixed formulations have a related constant-current gap, plus a separate results-layer bug in `rect_finalize_bus_injections!` that I found while scoping this — both written up in a comment on #431 rather than changed here.

### DIFF
--- a/src/state_indexing_helpers.jl
+++ b/src/state_indexing_helpers.jl
@@ -44,7 +44,13 @@ function area_tail_offset(data::ACPowerFlowData, dcn::DCNetwork)
     return 2 * n_buses + 4 * size(data.lcc.p_set, 1) + vsc_tail_length(dcn)
 end
 
-"""Update state vector based on values of fields of data."""
+"""Update state vector based on values of fields of data.
+
+The REF/PV power slots hold the *net* injection `P_gen - P_load_total`, where
+`P_load_total` includes the constant-current and constant-impedance ZIP terms evaluated at
+the bus's current voltage magnitude (see [`get_bus_active_power_total_withdrawals`](@ref)) —
+matching how `ACPowerFlowResidual` forms `P_net`/`Q_net`. Inverse of [`update_data!`](@ref).
+"""
 function update_state!(x::Vector{Float64},
     data::ACPowerFlowData,
     time_step::Int64,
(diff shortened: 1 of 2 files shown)


jd-lara (approved):

jd-lara:
@luke-kiernan we need to move this to psy6 branch too.

PR #441, opened 2026-09-01
test: cover flow reporting for nested reduction aggregates

luke-kiernan:
## What

Adds `test/test_nested_reduction_flow_reporting.jl`, covering branch-flow reporting for reduction aggregates that contain other reduction aggregates. One shape works and is now pinned; two are `@test_broken`.

## Background

`DegreeTwoReduction` deliberately builds two nesting shapes (PNM `degree_two_reduction.jl`, and PNM's own `test_nested_reduction_aggregates.jl`), and they compose:

| shape | container | member | reporting |
|---|---|---|---|
| parallel group as a chain segment | `BranchesSeries` | `BranchesParallel` | ✅ works |
| sibling chains in parallel | `BranchesParallel{BranchesSeries}` | `BranchesSeries` | ❌ broken |

In the places where PF expands the aggregate arc back to its physical branches (`_compute_segment_flows`, `_distribute_arc_flows`), we hand every member of a parallel group straight to the leaf-branch routine, so a chain nested inside a parallel group is never descended into. PNM itself *is* fully recursive here (`_subset_two_port`), so this is a PowerFlows-side gap.

Aside: because `AbstractReductionAggregate <: PSY.ACTransmission`, the chain satisfies those leaf signatures. So the failure lands deep in PSY, instead of getting a clean error in PF.

## What the tests do

Three minimal 3–4 bus fixtures, one per shape (buses 1 and 2 carry injections so they are irreducible endpoints; buses 3 and 4 are bare chain interiors).

- **Working shape** — passing tests: every physical branch appears in `BRANCH_FLOWS`, flows match an unreduced solve, and the chain's interior bus voltage is recovered on write-back. Covers the DC and PTDF paths too, which the existing coverage in `test_post_processing.jl` does not exercise.
- **Broken shapes** — `@test_broken` on the AC, DC, PTDF and write-back paths, plus a plain `@test` documenting that arc-level reporting collapses the nest to one row.

Each testset pins its fixture's aggregate shape with plain `@test`s, so if PNM's grouping changes the fixture stops silently testing the wrong thing.

## Why documented rather than fixed

Real datasets rarely nest this far. Under `DegreeTwoReduction`, ACTIVSg2000 has 410 parallel groups, **none** of them parallel-of-chains, and 7 chains with a parallel segment; `case10_radial_series_reductions` has 2 chains with a parallel segment and no parallel groups. The existing suite therefore cannot structurally reach the broken path.

The area already carries a standing caveat at `PowerFlowData.jl:569`: *"Degree 2 network reductions mis-report branch power flows, but bus voltage results are correct."*

## Test plan

```
julia --project=test -e 'using PowerFlows; include("test/PowerFlowsTests.jl"); using .PowerFlowsTests, ReTest; retest(PowerFlowsTests, r"nested aggregates")'
```

Result: **27 passed, 10 broken, 0 failed, 0 errored.** Formatter run; no other files touched.

## Adjacent, not addressed here

`_compute_segment_flows`'s `BranchesSeries` method assumes every member of a nested parallel segment shares the group's arc orientation, but `_build_chain_merge_group` exists precisely to hold anti-parallel members on a chain segment. An anti-frame member would get its endpoint voltages swapped. Not reproduced, so not tested here.

https://claude.ai/code/session_01RmjCvHjD7AfJcF5yccAndf

### DIFF
--- a/test/test_nested_reduction_flow_reporting.jl
+++ b/test/test_nested_reduction_flow_reporting.jl
@@ -1,0 +1,254 @@
+#=
+Flow reporting for reduction aggregates that contain other reduction aggregates.
+
+`DegreeTwoReduction` deliberately builds two nested shapes (PNM `degree_two_reduction.jl`,
+and PNM's own `test_nested_reduction_aggregates.jl`):
+
+  - a parallel group as a chain segment -- `BranchesSeries` holding a `BranchesParallel`
+  - sibling chains in parallel          -- `BranchesParallel{BranchesSeries}`
+
+and the two compose, giving a `BranchesParallel{BranchesSeries}` whose chains themselves
+carry a parallel segment.
+
+PowerFlows expands an aggregate arc back to its physical branches in two places, and both
+descend only into the first shape:
+
+  - AC:  `_compute_segment_flows` (`post_processing.jl`), which special-cases
+    `segment isa AbstractBranchesParallel` inside the `BranchesSeries` method but hands every
+    member of a `AbstractBranchesParallel` straight to `_segment_flow_entry`.
+  - DC/PTDF: `_distribute_arc_flows`, with the same asymmetry.
+
+So a chain nested inside a parallel group is never descended into. Because
+`AbstractReductionAggregate <: PSY.ACTransmission` (PNM `definitions.jl`), the chain is
+accepted by the leaf-branch method signatures rather than rejected at the PowerFlows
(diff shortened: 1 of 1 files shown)


jd-lara (commented):
make sure the CI/CD passes with the updated PNM

jd-lara (approved):

PR #443, opened 2026-09-04
feat: update calculation of reactive power losses on line components

mcllerena:
This PR update is based on the Qloss calculation description on PSSE's POM:

> Line losses are calculated as I^2.R and I^2.X, excluding line charging, line connected shunt, and transformer magnetizing admittance components.

Based on this, what we used to have on Qloss calculation was including the net reactive contribution of the shunt and charging elements.

Resulting in
image

With the change, we calculate the power associated with the series branch separately, so that we only use it's contribution on the Qloss, excluding line charging and line-connected shunts.

Resulting in 
image

This is based on the LL EI pf case currently being tested.

### DIFF
--- a/src/post_processing.jl
+++ b/src/post_processing.jl
@@ -477,7 +477,9 @@ end
     _segment_flow_entry(segment, V_from, V_to)
 
 Compute a `BranchFlowEntry` for a single segment given its endpoint voltages. Returns the
-from-to and to-from complex power flows, plus losses.
+from-to and to-from complex power flows, plus losses. For non-line segments, the
+reactive loss is the sum of the terminal reactive powers from the complete branch
+admittance model.
 """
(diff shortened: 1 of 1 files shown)


luke-kiernan (commented):
Code looks fine. But I question: do we want to match PSS(E) here? Is this PSS(E) being silly, or standard industry practice? Excluding those elements from the reactive loss calculation for lines alone is a strange inconsistency. I'd think of instead making the feature opt-in, or at least adding a flag to opt out.

edit: I suspect the POM probably meant line in the generic sense of "transmission elements." The implementation here takes line in the narrow sense, `Line` structs alone. Could check what PSS(E) does empirically to verify

jd-lara (approved):
The fix looks good to me

Issue #416, opened 2026-07-19
Parse PowerFlow settings from industrial tools


jd-lara:
It would be good to be able to collect all the settings from industrial tools to parse them and also export the cases with adequate parity

Issue #399, opened 2026-06-22
Export PowerFlow options with the PSSE exporter


jd-lara:
Currently we only export the data on the PowerFlow in the loop, we should also export the settings if there was an actively used solve call. 

> @luke-kiernan I think some of thse should actually be consistent with the PowerFlows algorithm used but that is another effort  

 _Originally posted by @jd-lara in [#397](https://github.com/Sienna-Platform/PowerFlows.jl/pull/397/changes#r3456124495)_

PR #446, opened 2026-09-10
SolutionParameters interface, PSS/E solution records, and discrete-control sensitivity dispatch

jd-lara:
Three commits on top of `psy6` at 756a827.

## 1. `SolutionParameters` as the parameter interface (closes #399, closes #416)

Solve parameters lived in three places at once: typed fields on each AC model, an untyped `solver_settings::Dict{Symbol,Any}` on the same models, and call-site keywords like `tol` and `maxIterations` that were never stored at all.

`SolutionParameters` replaces the dictionary and absorbs the six control flags (`check_reactive_power_limits`, `enhanced_flat_start`, `control_discrete_devices`, `area_interchange_control`, `interchange_tolerance`, `tie_definition`). Its field names match the keywords the solvers already accept, so `get_solver_kwargs` returns a NamedTuple built from the struct and no solver signature moves; a call-site keyword still overrides the stored value.

The same type is what a PSS/E v35 case file is read into and written from:

- **Export (#399).** Every v35 export previously wrote a frozen literal claiming tap adjustment, switched-shunt adjustment and area interchange control were all off — wrong whenever the solve used them, so the case reloaded as a different problem. The records are now generated from the solve the exporter is given through `update_exporter!`. With no solve attached the output is byte-identical to before, pinned by a golden test. The v33 raw format has no system-wide data section, so a v33 export warns and drops the parameters.
- **Import (#416).** `read_solution_parameters` fills a `SolutionParameters` from a case file, ready to hand back to an evaluation model. It returns `nothing` when a file carries no parameters rather than defaults that would look parsed.

The vendor format is confined to `psse_solution_records.jl`; no format name appears in the type, its fields, or the public functions.

`solver_settings` and the legacy keywords keep working — the dictionary raises a `depwarn`, the named keywords remain a supported spelling — so no construction site needed an edit.

## 2. Fix the marginal LCC multiperiod parity assertion

`LCC + control multiperiod: per-step equivalence` failed at a hair's breadth, and not because
the solutions disagreed. `isapprox(x, y; atol)` on two vectors compares `norm(x - y)`, a 2-norm
over the whole difference, and passing `atol` sets `rtol = 0`. Every bus was individually inside
the tolerance (worst 7.2e-8 against 1e-7); only the aggregation tripped it — and that also meant
the check silently tightened by `sqrt(n)` as a fixture grew. Forty lines away in the same file,
`_assert_steps_separated` already applied the same constant element-wise.

Parity now goes through a `_parity` helper built on the `all(isapprox.(...))` idiom the file already used, and `LCC_PARITY_ATOL` moves 1e-7 → 1e-6, matching `test_multiperiod_discrete_control.jl` (the same comparison without LCC converter state) and the inline 1e-6 already used further down this file. The constant is squeezed from both sides, so its comment now records the measured window — including that magnitude separation at 1.8e-6 is the
real ceiling.

This unblocked 29 previously-hidden tests: the suite went from 5419 to 5448 total.

## 3. Dispatch the discrete-control sensitivity layer on the formulation

Groundwork, no behaviour change. `_sensitivity_context` dispatched on `ACPolarPowerFlow` and the generic method returned `nothing`, so rectangular-CI and MCPB fell back to finite-difference probes and — via `use_batched = !isnothing(ctx)` — to a different pass schedule entirely. Three seams now separate what varies by formulation from what does not: `_sensitivity_residual_jacobian`, `_dF_dp!` (dispatching on the residual type), and
`_dVm_from_sol`.

Polar's `∂F/∂p` bodies are moved verbatim rather than unified behind a shared current-injection primitive: the tap would be bit-identical either way, but the shunt's `−Vm²` shortcut differs from `imag(V·conj(jV))` by 1 ulp in about a third of sample points, and the gain sizes the step rather than only its direction. `use_batched` is gated on a new `_refreshable` predicate rather than on the presence of a context, because rectangular and MCPB cache p-dependent data the polar refresh does not touch, and batching them before that refresh exists would be right on the first pass and silently wrong on every one after.

Polar's analytic gain is pinned bit-for-bit by a new test using `===`.

## Known state

The suite is at **5461 pass / 1 fail**. The one failure, `LCC + control: non-polar formulation (ACRectangularPowerFlow)`, is pre-existing — verified against the unmodified file — and was invisible until commit 2 unblocked it. It is not a tolerance problem: bus 101 differs by 5.8e-4 against a 1e-6 tolerance because the switched shunt settles one discrete step apart between formulations (polar 1.0, rectangular 1.5), which follows from the pass-schedule divergence commit 3 begins to close. Remaining work is tracked in the plan for this effort and is not part of this PR.

https://claude.ai/code/session_01MYUCoYTQBe4LKVdTHfcuj8

---
_Generated by [Claude Code](https://claude.ai/code/session_01MYUCoYTQBe4LKVdTHfcuj8)_


src/discrete_control/control_sensitivity.jl
```diff
@@ -0,0 +1,337 @@
+# Analytic sensitivity for the discrete-control continuation: dy/dp from the factored power
```
luke-kiernan, line 1: Boy, 20 lines of comments.

I'd cut lines 4-7 (details about include order).

> Deliberately NOT `initialize_power_flow_variables`

I'm inferring the comment applies to the `x = calculate_x0(data, ts)` line. Condense these 4 lines of comments to one line and move to before that line.

src/PowerFlows.jl
```diff
@@ -65,6 +67,9 @@ include("definitions.jl")
 # Before PowerFlowData.jl: defines PFLinearSolverCache and AbstractNRCache, which
 # type the lazily-populated cache slots on PowerFlowData.
 include("linear_solver_backend.jl")
+# Before `power_flow_types.jl`: the AC evaluation models embed a `SolutionParameters`.
```
luke-kiernan, line 70: more silly comments

src/psse_solution_records.jl
```diff
@@ -0,0 +1,475 @@
...
+
+"""
+Render a `Float64` without scientific notation, so the emitted records stay in the plain
+decimal form the format uses. `string(1e-5)` is `"1.0e-5"`, which would silently reformat
+records that are otherwise passed through unchanged.
+
+Falls back to `string` for exponents large enough that a decimal expansion would be
+absurd; the format accepts scientific notation there.
+"""
+function _decimal_string(x::Float64)
```
luke-kiernan, line 60: Briefly check if there's some built-in alternative or one in an existing dependency.

src/psse_solution_records.jl
```diff
@@ -0,0 +1,475 @@
...
+        length(parts) == 2 || continue
+        out[uppercase(strip(parts[1]))] = strip(parts[2])
+    end
+    return out
+end
+
+_record_int(d, key, fallback) =
+    haskey(d, key) ? something(tryparse(Int, d[key]), fallback) : fallback
+_record_float(d, key, fallback) =
+    haskey(d, key) ? something(tryparse(Float64, d[key]), fallback) : fallback
```
luke-kiernan, line 286: There's the default-value `get`...or are you reaching for `haskey` to avoid its eager behavior?

src/psse_solution_records.jl
```diff
@@ -0,0 +1,475 @@
...
+
+The mapped solution-record fields of a case file, or `nothing` when the file carries none
+— a format revision below 35, or a revision 35 file whose block is absent. `nothing`
+rather than defaults: defaults would be indistinguishable from parsed values.
+
+Unrecognized records and unrecognized field names are ignored, so a case written by a
+newer PSS/E than this mapping knows about still reads.
+"""
+function read_solution_records(path::AbstractString)
+    records = _significant_records(readlines(path))
```
luke-kiernan, line 325: > readlines(path)

Do we care about performance for large files? If so, switch to a read-one-process-one strategy instead of eagerly reading the whole file.

src/psse_solution_records.jl
```diff
@@ -0,0 +1,475 @@
...
+"""
+    SolutionRecordValues
+
+The subset of the solution records PowerFlows maps, in record units. Sits between
+[`SolutionParameters`](@ref) and the file so the writer and the reader share one
+description of what is mapped.
+
+`toln` is a mismatch in MW/MVAr, where the PowerFlows `tol` it comes from is an ∞-norm on
+the per-unit mismatch; the two differ by the system base.
+"""
```
luke-kiernan, line 89: I'd suggest including a 1-sentence description for each of these, as what to what they represent

test/test_discrete_control.jl
```diff
@@ -1059,6 +1061,102 @@ end
...
+                    # three Jacobians differ structurally, so the linear solve rounds
+                    # differently; measured worst case is ~6e-11.
+                    @test isapprox(g, reference[name]; rtol = 1e-6)
+                end
+            end
+        end
+    end
+end
+
+@testset "discrete control: polar analytic gain is pinned bit-for-bit" begin
```
luke-kiernan, line 1130: > any movement here means a refactor changed polar's arithmetic, which is exactly what must not happen.

I don't follow: why would it be a problem for the value to change by e.g. 1-ulp? Also, the hard-coded `-1.037005876824872` in the test is a magic number

test/test_solution_parameters.jl
```diff
@@ -0,0 +1,124 @@
+@testset "SolutionParameters defaults match the solver defaults" begin
```
luke-kiernan, line 1: This first test feels unnecessary, more like testing for typos than testing our code. Same throughout much of this file

luke-kiernan (commented):
First pass

luke-kiernan (commented):
Looks simple enough. There's a few things I commented on last time that are still there, but nothing so big I'd hold up merging over it

src/psse_solution_records.jl
jd-lara: I cut down the size of the function but this is still the best approach to cut numerical noise

src/psse_solution_records.jl
jd-lara: Done

jd-lara (commented):
We need to wire using the solution parameters into the POM interface afterwards

src/discrete_control/control_continuation.jl
```diff
@@ -910,7 +907,7 @@ which for a three-winding winding is suffixed per circuit. Non-transformer famil
 `device_name`/`circuit_index` `missing`.
 
 For a single-time-step solve (`time_steps == 1`), the solved settings are also written back
-to the `PSY.System` by [`solve_and_store_power_flow!`](@ref) under active controls, and
+to the `PSY.System` by `solve_and_store_power_flow!` under active controls, and
```
luke-kiernan: nitpick: removed `@ref`. Why?

test/Project.toml
```diff
@@ -46,13 +38,6 @@ PowerFlowFileParser = {url = "https://github.com/Sienna-Platform/PowerFlowFilePa
 PowerNetworkMatrices = {url = "https://github.com/Sienna-Platform/PowerNetworkMatrices.jl.git", rev = "psy6"}
 # PSY's own [sources] pins for these are ignored once PSY is a dependency rather than the
 # root project; the test env needs its own pins so Pkg can resolve PSY's unregistered OpenAPI deps.
-PowerCoreOpenAPIModels = {url = "https://github.com/Sienna-Platform/PowerOpenAPIModels.git", rev = "main", subdir = "PowerCoreOpenAPIModels.jl"}
```
luke-kiernan, line 49: Why is this no longer needed? Did you combine the `FooOpenAPI` mdoules into one monorepo, instead of keeping ~6 small ones?

luke-kiernan (approved):

test/Project.toml
jd-lara: It was split in two

src/discrete_control/control_continuation.jl
jd-lara: It was causing a failure, I can add it again

